PR #1, opened 2020-04-09
Add NativeCompiler stub

vchuravy:


vchuravy:
pairs with https://github.com/vchuravy/MCAnalyzer.jl/pull/18

I suspect I will need to add some more tests here.

PR #34, opened 2020-05-18
Fix AMDGPUnative reverse CI

jpsamaroo:


jpsamaroo:
Rebased on #71 and dropped `sink(i)` changes to see if the new `fix_alloca_addrspaces!` pass will rewrite it correctly, so I don't have to modify `sink(i)`.

PR #267, opened 2021-12-03
fix verifier warnings

aviatesk:
In order to handle callsite inlining correctly, now Julia's native
optimizer and verifier expects `src.ssaflags` to have the same length
as `src.code`.


vchuravy:
Thanks!

PR #272, opened 2021-12-17
Allow caller of classify_argument to specify that the first arg is sret

vchuravy:
`classify_arguments` uses the function type so it can't do that
determination on it's own.

cc: @wsmoses

### DIFF
--- a/src/irgen.jl
+++ b/src/irgen.jl
@@ -287,14 +287,14 @@ end
     GHOST       # not passed
 end
 
-function classify_arguments(@nospecialize(job::CompilerJob), codegen_ft::LLVM.FunctionType)
+function classify_arguments(@nospecialize(job::CompilerJob), codegen_ft::LLVM.FunctionType, has_sret=false)
     source_sig = Base.signature_type(job.source.f, job.source.tt)::Type
     source_types = [source_sig.parameters...]
 
     codegen_types = parameters(codegen_ft)
 
     args = []
-    codegen_i = 1
+    codegen_i = 1 + has_sret
     for (source_i, source_typ) in enumerate(source_types)
         if isghosttype(source_typ) || Core.Compiler.isconstType(source_typ)
             push!(args, (cc=GHOST, typ=source_typ))
--- a/test/native.jl
+++ b/test/native.jl
@@ -361,4 +361,22 @@ end
 
 ############################################################################################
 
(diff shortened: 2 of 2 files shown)


wsmoses:
Also sret as an attribute is not always set even if Julia makes an sret.

vchuravy:
```
julia> function meta(a, out, cond)
           if cond
               out[] = a
           end
       end
meta (generic function with 1 method)

julia> @code_llvm meta(2.0, Ref(0.0), true)
;  @ REPL[1]:1 within `meta`
define { {}*, i8 } @julia_meta_141([8 x i8]* noalias nocapture align 8 dereferenceable(8) %0, double %1, {}* nonnull align 8 dereferenceable(8) %2, i8 zeroext %3) #0 {
top:
;  @ REPL[1]:2 within `meta`
  %4 = and i8 %3, 1
  %.not = icmp eq i8 %4, 0
  br i1 %.not, label %L4, label %L2

L2:                                               ; preds = %top
;  @ REPL[1]:3 within `meta`
; ┌ @ refvalue.jl:57 within `setindex!`
; │┌ @ Base.jl:43 within `setproperty!`
    %5 = bitcast {}* %2 to double*
    store double %1, double* %5, align 8
; └└
  %.0..sroa_cast = bitcast [8 x i8]* %0 to double*
  store double %1, double* %.0..sroa_cast, align 8
  br label %L4

L4:                                               ; preds = %L2, %top
  %merge = phi { {}*, i8 } [ { {}* null, i8 1 }, %top ], [ { {}* null, i8 2 }, %L2 ]
  ret { {}*, i8 } %merge
}
```

maleadt:
We could pass it the function instead, but I guess that first need a fix in Julia to set the `sret` attribute.

wsmoses:
Only relevant for code calling the function in this PR, I remember @vtjnash said there was a compile-time performance reason why sret wasn't set and it didn't matter for correctness so I'm not sure there will be a proper Julia fix so much as a "derive it in the same way".

maleadt:
Can we easily derive it? Otherwise that's be a great thing for `classify_arguments` to do.

vtjnash:
There are also sometimes 2 sret arguments. In codegen we call `struct jl_returninfo_t get_specsig_function()` to re-compute this information precisely.

PR #282, opened 2022-02-06
Generalize classify arguments function

wsmoses:
This PR extends the classify_arguments API to take the signature type rather than the compiler job. This allows argument classification without having the runtime value of the function and enables the use of the utility in broader contexts (like Enzyme.jl)


maleadt (approved):

maleadt:
Stale.

PR #341, opened 2022-06-01
Remove Hack for GPUCompilation

collinwarner:
- Integrate ThreadSafe modules
- Now uses same AS throughout (no changing for Julia to support it)


collinwarner:
Making new branch since TSM are already merged in on master

PR #355, opened 2022-08-23
GCN: Set inline cost threshold to max

jpsamaroo:


jpsamaroo:
This doesn't appear to be necessary for what I needed it for, so closing for now

PR #359, opened 2022-10-19
Annotate exit/trap with noreturn & inaccessiblememonly

vchuravy:
This should help Enzyme to not assume that this assembly has
any effects it needs to handle.


maleadt (approved):
LGTM

PR #367, opened 2022-10-28
MethodInstances are unique so we can use an IdDict here.

vchuravy:
MIs are considered unique by pointer-identity. So use an IdDict here since that will allow for serialization and de-serialization.

cc: @collinwarner


maleadt (approved):
LGTM

PR #425, opened 2023-04-09
add GPUCompiler precompilation caching

collinwarner:
Adds ability to precompile code in GPUCompiler.GLOBAL_CI_CACHES. Taps into non-gpu caching of global constants to write the current instance of the global cache and reload on initialization. Requires user to declare, initialize and snapshot local cache. The user will then use GPUCompiler.precompile_gpucompiler. Mainly this adds and api for downstream packages such as Enzyme, CUDA, to use to cache instances of their functions. A sample SimpleGPU and Example.jl illustrate usage.


maleadt:
You forgot to commit `precompile_native.jl`.

maleadt:
Could you explain what the purpose/design of this PR is? It's not at all clear to me, and looking downstream lots of functionality is entirely unused (e.g. `reinit_cache`).

collinwarner:
> Could you explain what the purpose/design of this PR is? It's not at all clear to me, and looking downstream lots of functionality is entirely unused (e.g. `reinit_cache`).
>  
> I'm not sure why this needs anything in GPUCompiler.jl at all. Shouldn't it be sufficient for downstream packages to trigger a compilation to cache whatever they need, e.g., how JET.jl does it https://github.com/aviatesk/JET.jl/blob/b688eda6eb50a18e9e218d32650d2de23f085d50/src/JET.jl#L1382-L1396

Updated initial comment and added some example code. Hope this clears some things up!

maleadt:
Not really, sorry. Could you describe what's the problem you want to solve, why it doesn't work with current precompilation tools, and why you opted for the design you did? Those global undocumented macros (doing questionable things) are a very non-Julian API.

maleadt:
> The main issue is that GPUCompiler's GLOBAL_CI_CACHES are not persistent on reruns.

Why not? It's just a global dict, why doesn't it get serialized in the .ji file?

collinwarner:
> > The main issue is that GPUCompiler's GLOBAL_CI_CACHES are not persistent on reruns.
> 
> 
> 
> Why not? It's just a global dict, why doesn't it get serialized in the .ji file?


It is serialized, it just occurs to early in the process. By the time the dependent packages have inserted into the cache it is too late for the global. Additionally multiple children are now allowed to mutate and still have some see a cache improvements.

maleadt:
> It is serialized, it just occurs to early in the process.

Repeating my comment from Slack: Is this because the global is serialized as part of the GPUCompiler.ji, and isn't part of, e.g., CUDA.jl's precompilation image? In that case, you could override `ci_cache` and use a Dict that's serialized as part of the downstream package, in order to avoid this complexity.

If that turns out to be the way to do it, we could even remove the global CI cache here to force users to bring their own (and thus get proper precompilation of GPUCompiler-inferred CIs).

vchuravy:
So the overarching design consideration is:

Users of Enzyme.jl/CUDA.jl/AMDGPU.jl should be able to "precompile" their code.
Where can we store these precompilation results, while also ensuring that they get invalidated properly?

Each user package will need to declare an "anchor"/cache that will be serialized into the `.ji` of this package.
So the workflow is something like:

```
module ParallelStencil

using CUDA
using Enzyme
import GPUCompiler

GPUCompiler.@declare_cache() # anchor

f(x) = 1
CUDA.precompile_cuda(f, (Int, ))
Enzyme.precompile_fwd(f, (Int, ))

function __init__()
    GPUCompiler.@reinit_cache()
end

GPUCompiler.@snapshot_cache()
```

So it is not the down-stream packages of GPUCompiler that need to bring their own cache,
but it is the user of those packages.

We use the cachefile of `ParallelStencil` to save the cache entries that were discovered during precompilation of PS,
and we then need to re-insert those cache entries into the cache.

That's at least the high-level design Collin and I came up with.

src/precompile_native.jl
```diff
@@ -0,0 +1,115 @@
...
+    end
+end
+
+function get_code_cache_i(i)
+    for (j, cc) in enumerate(GPUCompiler.GLOBAL_CI_CACHES)
+        if j == i
+            return cc
+        end
+    end
+end
```
vchuravy, line 57:
```suggestion
```

vchuravy: This function has no use? no?

vchuravy (commented):
The new content in `examples` should likely go into `test`.

src/precompile_native.jl
collinwarner: Just a debugging function, removed

src/jlgen.jl
```diff
@@ -254,8 +254,28 @@ struct CodeCache
     dict::IdDict{MethodInstance,Vector{CodeInstance}}
 
     CodeCache() = new(Dict{MethodInstance,Vector{CodeInstance}}())
+    CodeCache(cache::CodeCache) = new(GPUCompiler.copyAndFilter(cache.dict))
+end
+
+function copyAndFilter(dict::IdDict)
```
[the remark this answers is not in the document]
collinwarner: That is used in https://github.com/collinwarner/GPUCompiler.jl/blob/3dbe9d5b7c7c5f56f18553f0e4d4bd9c2bdcaca5/src/precompile_native.jl#L102

It creates a CodeCache that contains unbounded entries only. Used when snapshotting.

src/precompile_native.jl
```diff
@@ -0,0 +1,115 @@
+const CACHE_NAME = gensym(:CACHE) # is now a const symbol (not a variable)
```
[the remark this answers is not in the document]
collinwarner: Done

src/precompile_native.jl
```diff
@@ -0,0 +1,115 @@
...
+is_precompiling() = ccall(:jl_generating_output, Cint, ()) != 0
+
+struct NativeCompilerParams <: AbstractCompilerParams end
```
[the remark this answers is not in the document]
collinwarner: Done

maleadt:
Why does the API consist of macros? Why doesn't something like this work:

```julia
module DownstreamPackage

using GPUCompiler, CUDA

const cache_snapshot = GPUCompiler.ci_cache_snapshot()
include("precompile.jl")
const cache = GPUCompiler.ci_cache_delta(cache_snapshot)

__init__() = GPUCompiler.ci_cache_insert(cache)

end
```

collinwarner:
> Why does the API consist of macros? Why doesn't something like this work:
> 
> ```julia
> module DownstreamPackage
> 
> using GPUCompiler, CUDA
> 
> const cache_snapshot = GPUCompiler.ci_cache_snapshot()
> include("precompile.jl")
> const cache = GPUCompiler.ci_cache_delta(cache_snapshot)
> 
> __init__() = GPUCompiler.ci_cache_insert(cache)
> 
> end
> ```

That would seem to work. You would then be doing working of needing to parse a delta vs just adding in what the local cache has created. But it would remove the macros

collinwarner:
Changed API to follow @maleadt advice. Leads to a cleaner interface. Added an example kernel with caching at test/ExamplePersistentCache/GPUKernel.jl . Using this you are able to get a persistent cache, which reduces the recompilation time on consecutive loads.

collinwarner:
Remaining work is to test integration with Downstream packages such as Enzyme, Oceananigans, CUDA, AMDGPU,.... Additionally, there are potentially some algorithmic improvements for the merge algorithm to bring precompile times with and without this feature more inline.

collinwarner:
image

src/precompilation_cache.jl
```diff
@@ -0,0 +1,186 @@
...
+    if GPUCompiler.ci_cache_lookup(cache, job.source, job.world, typemax(Cint)) === nothing
+        GPUCompiler.ci_cache_populate(interp, cache, mt, job.source, job.world, typemax(Cint))
+    end
+end
+
+"""
+Reloads Global Cache from global variable which stores the previous
+cached results
+"""
+function reinit_cache(LOCAL_CACHE)
```
maleadt, line 133: What is this used for?

maleadt:
cc @aviatesk, this may be relevant to DAECompiler (as a workaround, until we have the ability to update another module's globals, i.e., a ci cache).

src/precompilation_cache.jl
collinwarner: oops reminent code

collinwarner:
See greater performance improvement when used during Enzyme.jl's precompilation phase. [https://github.com/EnzymeAD/Enzyme.jl/pull/760](url)
image

src/precompilation_cache.jl
collinwarner: Dead code, removed

src/jlgen.jl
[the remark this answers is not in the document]
vchuravy: can we just write this as `filter(validate_codecache, cache.dict)` where valid is:

```
function validate_codecache(cache)
    for ci in cache
        if ci.max_world < typemax(typeof(ci.max_world))
            return false
        end
        return true
    end
end
```

But that seems overeager, are we gurantueed just one entry? Or do we want to remove all CIs that don't have max_world?

src/precompilation_cache.jl
```diff
@@ -0,0 +1,120 @@
+const CACHE_NAME = gensym(:CACHE) # is now a const symbol (not a variable)
+is_precompiling() = ccall(:jl_generating_output, Cint, ()) != 0
+
+export ci_cache_snapshot, ci_cache_delta, ci_cache_insert, precompile_gpucompiler
+
+function ci_cache_snapshot()
+    cleaned_cache_to_save = IdDict()
```
vchuravy, line 7: Is this just `copy(GPUCompiler.GLOBAL_CI_CACHES)`?

collinwarner: There is an additional parse when constructing the CodeCache that removes CodeInstances in finite ranges. I could potentially split up that process so there are two phases. Copying then filtering, I though since we were already doing one pass over the data we could add filtering in directly.

collinwarner:
Improves downstream CUDA code
image

PR #456, opened 2023-05-22
Add llvm_gnu_pubnames to Interface and set llvm_debug_info for native target

vchuravy:


src/native.jl
```diff
@@ -37,3 +37,13 @@ end
 
 runtime_slug(job::CompilerJob{NativeCompilerTarget}) = "native_$(job.config.target.cpu)-$(hash(job.config.target.features))$(job.config.target.jlruntime ? "-jlrt" : "")"
 uses_julia_runtime(job::CompilerJob{NativeCompilerTarget}) = job.config.target.jlruntime
+
+function llvm_debug_info(@nospecialize(job::CompilerJob{NativeCompilerTarget}))
```
vchuravy, line 41: @gbaraldi I was debating to use simply `FULL` or ` Base.default_debug_info_kind()` for all Native targets,
but maybe best not to change this for StaticCompiler.jl?

PR #494, opened 2023-08-16
Allow for generic MethodTableView

vchuravy:
Experiment for https://github.com/JuliaGPU/CUDA.jl/issues/1991 with https://github.com/vchuravy/FastmathOverlay.jl

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -177,7 +177,7 @@ isintrinsic(@nospecialize(job::CompilerJob), fn::String) = false
 
 # provide a specific interpreter to use.
 get_interpreter(@nospecialize(job::CompilerJob)) =
-    GPUInterpreter(job.world; method_table=method_table(job),
+    GPUInterpreter(job.world; method_table_view=maybe_cached(method_table_view(job, world)),
                    code_cache=ci_cache(job), inf_params=inference_params(job),
                    opt_params=optimization_params(job))
 
@@ -219,6 +219,7 @@ end
 
 # the method table to use
 method_table(@nospecialize(job::CompilerJob)) = GLOBAL_METHOD_TABLE
+method_table_view(@nospecialize(job::CompilerJob), world) = get_method_table_view(world, method_table(job))
 
 # the inference parameters to use when constructing the GPUInterpreter
 function inference_params(@nospecialize(job::CompilerJob))
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -324,34 +324,23 @@ end
 
 ## interpreter
 
-if isdefined(Base.Experimental, Symbol("@overlay"))
-    using Core.Compiler: OverlayMethodTable
-    const MTType = Core.MethodTable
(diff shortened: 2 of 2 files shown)


test/utils.jl
```diff
@@ -92,3 +92,75 @@ end
...
+@testset "StackedMethodTable -- Overlayed" begin
+    o_sin = findsup(Tuple{typeof(sin), Float64}, OverlayMT())
+    s_sin = findsup(Tuple{typeof(sin), Float64}, StackedMT())
+    ss_sin = findsup(Tuple{typeof(sin), Float64}, DoubleStackedMT())
+    @test s_sin == o_sin
+    @test ss_sin == o_sin
+
+    worlds = o_sin[2]
+    @test worlds.min_world == next_world
+    @test worlds.max_world == typemax(typeof(next_world))
```
vchuravy, line 151: On 1.12:

```
   Expression: worlds.min_world == next_world
   Evaluated: 0x00000000000097e8 == 0x00000000000097e9
```

This is probable the bindings change making this test more fragile...

aviatesk (commented):
From an implementation perspective, this seems fine.
One point to note is that in the current overlay method table mehcanism, simply looking at `.fully_covers` cannot reproduce the specialization behavior that occurs in normal dispatch.
Consider the following example:

```julia
Base.Experimental.@MethodTable mt

func(a::Number) = :number

func(a::Integer) = :integer

Base.Experimental.@overlay mt func(a::Number) = :number_overlay
```

In this case, the correct specialization for `Tuple{typeof(func), Int}` should naturally be dispatch to `func(::Integer)`:
```julia
julia> Base.Compiler.findall(Tuple{typeof(func),Int}, Base.Compiler.InternalMethodTable(Base.get_world_counter()))
Compiler.MethodLookupResult(Any[Core.MethodMatch(Tuple{typeof(func), Int64}, svec(), func(a::Integer) @ Main REPL[14]:1, true)], Compiler.WorldRange(0x00000000000097cd, 0xffffffffffffffff), false)
```

However, when using overlay method tables, this rule is not preserved, and something like "local specialization within the overlay method table" takes in place:
```julia
julia> result = Base.Compiler.findall(Tuple{typeof(func),Int}, Base.Compiler.OverlayMethodTable(Base.get_world_counter(), mt))
Compiler.MethodLookupResult(Any[Core.MethodMatch(Tuple{typeof(func), Int64}, svec(), func(a::Number) @ Main REPL[15]:1, true)], Compiler.WorldRange(0x00000000000097ce, 0xffffffffffffffff), false)

julia> result.matches[1].fully_covers
true
```

This issue likely applies to this PR as well. It might be worth being careful about whether there are any problems regarding this point. This is more of an implementation issue with `Base.Compiler.findall(type, ::OverlayMethodTable)` rather than a problem on the side that uses `OverlayMethodTable` though.

vchuravy:
> This issue likely applies to this PR as well. It might be worth being careful about whether there are any problems regarding this point. 

For our uses that is fine, but yes for some of the potential use-cases that might be an issue.

PR #508, opened 2023-09-07
Add a function to reset the compile cache

Seelengrab:
It can happen sometimes that a bad LLVM module makes it into the compile cache, which is really annoying to debug due to the very unhelpful LLVM error this causes (the `try`/`catch` [here](https://github.com/JuliaGPU/GPUCompiler.jl/blob/d9b8f4747cf789139fb2586ee67ced93967a8237/src/rtlib.jl#L125-L134) never fires, as LLVM seems to kill the process). The easiest fix is just to get a clean slate, by deleting the compile cache in the scratch space. This PR adds a function to do just that - it's already been useful in debugging a bad AVR runtime cache :)

### DIFF
--- a/src/GPUCompiler.jl
+++ b/src/GPUCompiler.jl
@@ -78,5 +78,7 @@ function reset_compile_cache()
 
     return
 end
+# backwards compatibility
+const reset_runtime = reset_compile_cache
 
 end # module
--- a/src/GPUCompiler.jl
+++ b/src/GPUCompiler.jl
@@ -51,7 +51,10 @@ compile_cache = "" # defined in __init__()
 
 function __init__()
     STDERR_HAS_COLOR[] = get(stderr, :color, false)
+    create_compile_cache()
+end
 
+function create_compile_cache()
     dir = @get_scratch!("compiled")
     ## add the Julia version
     dir = joinpath(dir, "v$(VERSION.major).$(VERSION.minor)")
@@ -62,6 +65,12 @@ function __init__()
     end
     mkpath(dir)
     global compile_cache = dir
+end
+
+function reset_compile_cache()
+    dir = @get_scratch!("compiled")
+    rm(dir; force=true, recursive=true)
+    create_compile_cache()
 end
 
 end # module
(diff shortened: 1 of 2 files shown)


Seelengrab:
Huh, just noticed this:

https://github.com/JuliaGPU/GPUCompiler.jl/blob/d9b8f4747cf789139fb2586ee67ced93967a8237/src/rtlib.jl#L152-L160

I suppose that's already the same thing, except it doesn't take care of the additional versioning scheme with 1.9+? Should I delete this or close this PR?

Seelengrab:
The failure on julia master seems to be due to https://github.com/JuliaGPU/GPUCompiler.jl/issues/506

maleadt:
> due to the very unhelpful LLVM error this causes (the `try`/`catch` [here](https://github.com/JuliaGPU/GPUCompiler.jl/blob/d9b8f4747cf789139fb2586ee67ced93967a8237/src/rtlib.jl#L125-L134) never fires, as LLVM seems to kill the process)

Yeah, I'm not sure why. When trying in isolation, reading invalid bitcode does throw a proper error. I'm not sure why it suddenly exits there sometimes...

> I suppose that's already the same thing, except it doesn't take care of the additional versioning scheme with 1.9+?

That may be a good thing to add.

Regarding the usefulness of this PR, I'm not sure it's worth the changes right now. Renaming `reset_runtime` is better, but seems like a unnecessary change given that we hope to remove this function at some point in the future (and support automatic invalidation, like Base does). You also shouldn't need to ever have to call `create_compile_cache()`, that should happen automatically (if it doesn't, that's a bug).

Seelengrab:
> You also shouldn't need to ever have to call create_compile_cache(), that should happen automatically (if it doesn't, that's a bug).

Right, that's why none of this is exported. It's just useful to have a function for this that can be called in both `__init__` and the reset of the compile cache.

There are also various places where `mkpath(compile_cache)` is called explicitly in the codebase, such as here:

https://github.com/JuliaGPU/GPUCompiler.jl/blob/d9b8f4747cf789139fb2586ee67ced93967a8237/src/rtlib.jl#L136-L139

which may be replaced with a call to `create_compile_cache` (with a check & bail inside) instead.

maleadt:
> Well, I can pretty reliably reproduce this with AVR, would you be interested in a MWE for that?

Yes, please!

PR #520, opened 2023-09-20
After conversion to LLVM we should be able to delete the inferred source of the kernel.

vchuravy:
@simonbyrne has shown me a heap-snapshot were the inferred source took up >>1GB of ram.

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -413,7 +413,7 @@ end
 function ci_cache_lookup(cache, mi, min_world, max_world; allow_uninferred=false)
     wvc = WorldView(cache, min_world, max_world)
     ci = CC.get(wvc, mi, nothing)
-    if ci !== nothing && (allow_uninferred || ci.inferred === nothing)
+    if ci !== nothing && (!allow_uninferred && ci.inferred === nothing)
         # if for some reason we did end up with a codeinfo without inferred source, e.g.,
         # because of calling `Base.return_types` which only sets rettyp, pretend we didn't
         # run inference so that we re-infer now and not during codegen (which is disallowed)
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -376,12 +376,6 @@ function CC.getindex(wvc::WorldView{CodeCache}, mi::MethodInstance)
 end
 
 function CC.setindex!(wvc::WorldView{CodeCache}, ci::CodeInstance, mi::MethodInstance)
-    src = if ci.inferred isa Vector{UInt8}
-        ccall(:jl_uncompress_ir, Any, (Any, Ptr{Cvoid}, Any),
-                mi.def, C_NULL, ci.inferred)
-    else
-        ci.inferred
-    end
     CC.setindex!(wvc.cache, ci, mi)
 end
 
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -615,7 +615,8 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
 
     if job.config.kernel
         # Don't cache the top-level inference result.
-        compiled[job.source].ci.inferred = nothing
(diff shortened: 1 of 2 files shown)


src/jlgen.jl
```diff
@@ -613,5 +613,10 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
         end
     end
 
+    if job.config.kernel
+        # Don't cache the top-level inference result.
+        compiled[job.source].ci.inferred = nothing
```
vchuravy, line 618:
```suggestion
        ci = compiled[job.source].ci
        Base.@atomic :release ci.inferred = nothing
```

src/jlgen.jl
```diff
@@ -613,5 +613,11 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
         end
     end
 
+    if job.config.kernel
+        # Don't cache the top-level inference result.
+        ci = compiled[job.source].ci
+        Base.@atomic :release ci.inferred = nothing
```
vchuravy, line 619:
```suggestion
        ci = compiled[job.source].ci
@static VERSION < v"1.9.0"
        ci.inferred = nothing
else
        Base.@atomic :release ci.inferred = nothing
end
```

simonbyrne:
This doesn't seem to fix my issue. I'm not sure exactly where the problem is, but I did notice:
```
julia> GPUCompiler.GLOBAL_CI_CACHES
Dict{CompilerConfig, GPUCompiler.CodeCache} with 2 entries:
  CompilerConfig for PTXCompilerTarget => CodeCache(IdDict{MethodInstance, Vector{CodeInstance}}(MethodInstance for >>(…
  CompilerConfig for PTXCompilerTarget => CodeCache(IdDict{MethodInstance, Vector{CodeInstance}}(MethodInstance for >>(…

julia> Base.summarysize(GPUCompiler.GLOBAL_CI_CACHES) / 10^6
1396.946174

julia> Base.summarysize(collect(values(GPUCompiler.GLOBAL_CI_CACHES))[1]) / 10^6
1393.855007

julia> Base.summarysize(collect(values(GPUCompiler.GLOBAL_CI_CACHES))[2]) / 10^6
3.090233
```

I tried manually calling `empty!` on this dict: it didn't seem to make any difference, so I suspect the data is being retaine somewhere else as well.

simonbyrne:
Also, what's odd is that RES reported by `top` is 6.3g, but
```
julia> Sys.maxrss() / 10^9
17.232601088
```

maleadt:
Removed a call to jl_uncompress_ir, as IIRC it was only needed for the 1.6 overlay hack: https://github.com/JuliaGPU/GPUCompiler.jl/pull/151#issuecomment-779687366
Maybe that also helps?

simonbyrne:
Unfortunately still no.

maleadt:
You could try taking a heap snapshot.

simonbyrne:
I did that: it looks like most of it is still the inferred objects:
Screenshot 2023-09-21 at 11 11 59 AM

I tried clearing them out manually: 
```
for cache in values(GPUCompiler.GLOBAL_CI_CACHES)
    for insts in values(cache.dict)
        for inst in insts
            @atomic :release inst.inferred = nothing
        end
    end
end
```
that seemed to work:

Screenshot 2023-09-21 at 11 09 45 AM

`top` is still reporting 4GB of memory usage though, so not sure what is going on.

vchuravy:
So I am only deleting top-level kernel calls. Since everything else is re-usable.

vchuravy:
@maleadt are we tracking anywhere how big the modules are we load onto the GPU?

maleadt:
> @maleadt are we tracking anywhere how big the modules are we load onto the GPU?

No, and I don't know of a way to query the size of a CuModule or CuContext.

PR #553, opened 2024-03-07
Specialize MI if necessary

vchuravy:
@vtjnash is there a way to speedup this query? For `identity(::Int)` this currently takes ~500ns

The goal here is to compute the `mi` we would have `invoked`.

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -69,8 +69,15 @@ function methodinstance(ft, tt, world=tls_world_age())
                (Any, Csize_t, Any),
                sig, world, #=method_table=# nothing)
     mi === nothing && throw(MethodError(ft, tt, world))
-
-    return mi::MethodInstance
+    mi = mi::MethodInstance
+    # `jl_method_lookup_by_tt` and `jl_method_lookup` can return a unspecialized mi
+    if mi.specTypes !== sig
+        # XXX: This slows down the lookup significantly
+        #      Do we really need this? We need the mi where we will insert our CodeInstance.
+        mi = CC.specialize_method(mi.def, sig, Core.svec(), preexisting=true)
+        mi === nothing && throw(MethodError(ft, tt, world))
+    end
+    return mi
 end
 
 # on older versions of Julia, the run-time lookup is much slower, so we'll need to cache it
@@ -82,7 +89,7 @@ function methodinstance(ft::Type, tt::Type, world::Integer)
     match, _ = CC._findsup(sig, nothing, world)
     match === nothing && throw(MethodError(ft, tt, world))
 
-    mi = CC.specialize_method(match)
+    mi = CC.specialize_method(match) # XXX: preexisting=true?
 
     return mi::MethodInstance
 end
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -14,7 +14,7 @@ tls_world_age() = ccall(:jl_get_tls_world_age, UInt, ())
 export methodinstance
 
 @inline function signature_type_by_tt(ft::Type, tt::Type)
-    u = Base.unwrap_unionall(tt)
+    u = Base.unwrap_unionall(tt)::DataType
     return Base.rewrap_unionall(Tuple{ft, u.parameters...}, tt)
 end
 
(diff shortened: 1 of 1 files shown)


src/jlgen.jl
```diff
@@ -62,15 +62,20 @@ methodinstance
...
 
     mi = ccall(:jl_method_lookup_by_tt, Any,
                (Any, Csize_t, Any),
                sig, world, #=method_table=# nothing)
     mi === nothing && throw(MethodError(ft, tt, world))
-
-    return mi::MethodInstance
+    mi = mi::MethodInstance
+    # `jl_method_lookup_by_tt` and `jl_method_lookup` can return a unspecialized mi
+    if mi.specTypes !== sig
```
vtjnash, line 74: fast path?
```suggestion
    if !isdispatchtuple(mi.specTypes) && sig != mi.specTypes
```

src/jlgen.jl
```diff
@@ -62,15 +62,20 @@ methodinstance
...
+        mi = CC.specialize_method(mi.def, sig, Core.svec(), preexisting=true)
+        mi === nothing && throw(MethodError(ft, tt, world))
```
vtjnash, line 76: AFAIK, the sparams is not optional here, unless you enjoy corrupt caches leading to memory corruption
```suggestion
        sparams = ...
        mi = CC.specialize_method(mi.def, sig, sparams)::MethodInstance
```
(someday I want it to be optional, and computed on-demand if empty, but that day is probably not today)

vchuravy: I was looking at uses of `specialize_method` in Base and seen a couple of places where it uses `sparams=Core.svec()`

I don't think I actually know  what `sparams` ought to be.

vchuravy: Could I use `mi.sparams_vals` e.g. reuse the sparams from the unspecialized mi?

vtjnash: yeah, that seems probably true

vchuravy:
So this is now ~70ns

vtjnash (commented):
`!isdispatchtuple(mi.specTypes)` is true, you are guaranteed to get back the exact same `mi` object

vchuravy:
Okay using `!==` instead of `!=` gets me the performance back.

```
julia> @benchmark GPUCompiler.methodinstance(typeof(identity), Tuple{Int,})
BenchmarkTools.Trial: 10000 samples with 968 evaluations.
 Range (min … max):  78.574 ns … 118.419 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     84.008 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   83.506 ns ±   2.955 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅                        █▄ ▂▂      ▂                         
  █▁▂▂▁▁▂▄▂▃▂▁▁▂▃▆▂▅▄▂▂▃▃▂▃██▂██▅▂▂▂▂▂█▂▄▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▃ ▃
  78.6 ns         Histogram: frequency by time         90.7 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark GPUCompiler.methodinstance(typeof(+), Tuple{Int,Int})
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  12.633 ns … 31.822 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     12.893 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   12.903 ns ±  0.353 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                 █             ▅ ▁             
  ▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▃█▄▁▂▁▁▁▁▃█▂▇▁▄█▂█▁▃▃▁▃▂▂▁▁▂▂ ▃
  12.6 ns         Histogram: frequency by time          13 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

vtjnash:
With `isdispatchtuple`, that should not be reachable

vtjnash:
Could I interest you in fixing this in JuliaInterpreter, now that you know all of the secrets?

vchuravy:
I could fix this in Base for `Base.method_instance`, but other than that I currently trying to not do code development and focus on my thesis

PR #567, opened 2024-04-12
Add a testsuite for inference caching

vchuravy:


test/precompile_tests.jl
```diff
@@ -0,0 +1,99 @@
...
+        end
+
+        let
+            job, _ = Native.create_job(kernel, ())
+            GPUCompiler.code_typed(job)
+        end
+        
+        # identity is foreign
+        # Maybe https://github.com/JuliaLang/julia/pull/49391
+        job, _ = Native.create_job(identity, (Int,))
```
vchuravy, line 68: @timholy can you remind me of the rules around caching of foreign code-instances,

I expected this to work, but maybe I need to dive into how these CodeInstances get queued.

src/jlgen.jl
```diff
@@ -749,3 +749,14 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
 
     return llvm_mod, compiled
 end
+
+function CC.typeinf(interp::GPUInterpreter, frame::CC.InferenceState)
```
vchuravy, line 753: Shouldn't do things without drinking coffee first. This should needs a version bound to limit it for 1.11.

In 1.11-beta2 it should not be necessary anymore

PR #586, opened 2024-05-30
Rework cache to use inline cache

vchuravy:
We currently put entries into the cache under two different keys.

- objectid(src), tls_world, cfg
- ci, cfg

The first is meant as a fast hit in scenarios where the current world-age is not changing.
If we follow Julia Base, we could instead replace the first with a linear search over a small
number of entries. 

By storing the ci as a field in the cache struct we quickly check if the cache is still valid,
currently when the `tls_world` changes we fall-back to the slower cache check.

I was thinking of adding something similar to Enzyme.
This is an API breakage though. So I am not convinced it is worth it right now.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -64,6 +64,36 @@ end
 
 const cache_lock = ReentrantLock()
 
+mutable struct CacheEntry
+    const ci::Core.CodeInstance
+    const obj::Any
+    @atomic next::Union{CacheEntry, Nothing}
+end
+
+@inline function lookup(entry::CacheEntry, world::UInt)::Any
+    while entry !== nothing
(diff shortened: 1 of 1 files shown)


src/execution.jl
```diff
@@ -80,65 +110,56 @@ session-dependent objects (e.g., a `CuModule`).
 function cached_compilation(cache::AbstractDict{<:Any,V},
```
vchuravy, line 110: This would be the API breakage. We need to change this to `AbstractDict{<:Any, CacheEntry}{V}}` or take `V` as a separate parameter.

vchuravy:
Stale with CompilerCaching.jl

PR #634, opened 2024-09-26
Make GPUInterpreter extensible

vchuravy:
Currently Enzyme uses it's own AbstractInterpreter, in particular to
handle inlining blocking of functions with custom rules and to handle
nested autodiff operations.

- [ ] Create a version of Enzyme with this
- [ ] Support a version of `gpuc.deferred(meta)`

PR #633, opened 2024-09-26
Add optimization callbacks that fire on a marker function

vchuravy:

PR #582, opened 2024-05-13
New deferred_codegen implementation

vchuravy:
Draft implementation https://github.com/JuliaGPU/GPUCompiler.jl/issues/581

@aviatesk pointed me to:
- https://github.com/JuliaLang/julia/blob/b9f68ac0afe7ad896fe2803b139b9c32103ac417/base/compiler/ssair/inlining.jl#L580
- https://github.com/JuliaLang/julia/blob/d61db204e576fa53bd897b2e6b1298ad9ad39f40/base/compiler/ssair/inlining.jl#L995-L996

For the refinement. So we will need a small pass that rewrites the call, post abstract interpretation.

PR #636, opened 2024-10-02
Mock Enzyme plugin

vchuravy:
Make sure that the new infrastructure can handle the complicate song and dance Enzyme needs to do.

### DIFF
--- a/test/plugin_testsetup.jl
+++ b/test/plugin_testsetup.jl
@@ -37,7 +37,7 @@ struct NeverInlineMeta <: InlineStateMeta end
 import GPUCompiler: abstract_call_known, GPUInterpreter
 import Core.Compiler: CallMeta, Effects, NoCallInfo, ArgInfo,
                       StmtInfo, AbsIntState, EFFECTS_TOTAL,
-                      MethodResultPure
+                      MethodResultPure, CallInfo, IRCode
 
 function abstract_call_known(meta::InlineStateMeta, interp::GPUInterpreter, @nospecialize(f),
                              arginfo::ArgInfo, si::StmtInfo, sv::AbsIntState, max_methods::Int)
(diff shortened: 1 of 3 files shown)


vchuravy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/JuliaGPU/GPUCompiler.jl/636?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#636** <a href="https://app.graphite.dev/github/pr/JuliaGPU/GPUCompiler.jl/636?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#634** <a href="https://app.graphite.dev/github/pr/JuliaGPU/GPUCompiler.jl/634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#633** <a href="https://app.graphite.dev/github/pr/JuliaGPU/GPUCompiler.jl/633?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#582** <a href="https://app.graphite.dev/github/pr/JuliaGPU/GPUCompiler.jl/582?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @vchuravy and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

test/plugin_testsetup.jl
```diff
@@ -69,5 +69,143 @@ function inlining_handler(meta::InlineStateMeta, interp::GPUInterpreter, @nospec
...
+        "extern gpuc.lookup",
+        Ptr{Cvoid},
+        Core.svec(#=meta=# Any, #=mi=# Any, #=f=# Any, primal_argtypes...), # Must use Any for MethodInstance or ftype
+        0,
+        QuoteNode(:llvmcall),
+        deferred_info.meta,
+        case.invoke,
+        primal_args...
+    )
+    ptr = insert_node!(ir, (idx += 1), NewInstruction(expr, Ptr{Cvoid}))
```
vchuravy, line 180: @aviatesk does this look correct?

test/ptx_tests.jl
```diff
@@ -504,4 +504,25 @@ end
...
+        unsafe_store!(a, y)
+        nothing
+    end
+    
+
+    @show PTX.code_typed(kernel, Tuple{Ptr{Float64}, Float64}, meta=Plugin.MockEnzymeMeta())
+
+    # FIXME: the fact that meta is necessary here almost invalidates that extension mechanism
+    #        we somehow need to be able to add this kind of "autodiff" abs int handling.
+    ir = sprint(io->PTX.code_llvm(io, kernel, Tuple{Ptr{Float64}, Float64}, meta=Plugin.MockEnzymeMeta()))
```
vchuravy, line 524: Grrmbl, I need to rethink #634 since I somehow need to tell CUDA version of GPUCompiler what to do when it sees a call to `autodiff`

test/plugin_testsetup.jl
```diff
@@ -69,5 +69,178 @@ function inlining_handler(meta::InlineStateMeta, interp::GPUInterpreter, @nospec
...
+            LLVM.position!(builder, call)
+            ops = LLVM.operands(call)
+            target = ops[1]
+            if target isa LLVM.ConstantExpr && (LLVM.opcode(target) == LLVM.API.LLVMPtrToInt ||
+                                                LLVM.opcode(target) == LLVM.API.LLVMBitCast)
+                target = first(LLVM.operands(target))
+            end
+            funcT = LLVM.called_type(call)
+            funcT = LLVM.FunctionType(LLVM.return_type(funcT), LLVM.parameters(funcT)[3:end])
+            direct_call =  LLVM.call!(builder, funcT, target, ops[3:end - 1]) # why is the -1 necessary
```
wsmoses, line 229: the last op is the called value

test/plugin_testsetup.jl
```diff
@@ -69,5 +69,178 @@ function inlining_handler(meta::InlineStateMeta, interp::GPUInterpreter, @nospec
...
+    # TODO: Effects
+    # aviatesk mentioned using inlining_policy instead...
+    itodo = Core.Compiler.InliningTodo(case.invoke, ir, Core.Compiler.Effects())
+    @assert itodo.linear_inline_eligible
+    push!(todo, (stmt_idx=>itodo))
+
+    return nothing
+end
+
+function mock_enzyme!(@nospecialize(job), intrinsic, mod::LLVM.Module)
```
vchuravy, line 214: Feedback from @wsmoses 

This pass would need access to the `compiled` dictionary so that Enzyme can do a lookup from "emitted function" to Julia function. This would speak for a different phase order in #633 

Or we add two callbacks "one early" and one "as part of the optimization pipeline"

test/plugin_testsetup.jl
```diff
@@ -69,5 +69,178 @@ function inlining_handler(meta::InlineStateMeta, interp::GPUInterpreter, @nospec
...
+
+function autodiff end
+
+import GPUCompiler: DeferredCallInfo
+struct AutodiffCallInfo <: CallInfo
+    rt
+    info::DeferredCallInfo
+end
+
+function abstract_call_known(meta::Nothing, interp::GPUInterpreter, f::typeof(autodiff),
```
aviatesk, line 88: `abstract_call_known` with this signature would never be called from `Core.Compiler`, so this overload would do nothing?

test/plugin_testsetup.jl
aviatesk: `handle_call!` isn’t meant to be overloaded, so I think this approach is preferred:
```julia
function Core.Compiler.src_inlining_policy(interp::GPUCompiler#=or EnzymeInterpreter?=#,
    @nospecialize(src), info::AutodiffCallInfo, stmt_flag::UInt32)
    # Goal:
    # The IR we want to return here is:
    # unpack the args ..
    # ptr = gpuc.deferred(MockEnzymeMeta(), f, primal_args...) 
    # ret = ccall("extern __autodiff", llvmcall, RT, Tuple{Ptr{Cvoid, args...}}, ptr, adjoint_args...)
    ir = Core.Compiler.IRCode() # contains a placeholder
    ...
    return ir
end
```
By overloading `src_inlining_policy` (or `inlining_policy` in older versions), we can apply this custom inlining to const-propped call sites and semi-concrete interpreted call sites as well.

vchuravy: Haha, then I have misunderstood the comment in:

https://github.com/JuliaLang/julia/blob/b9d9b69165493f6fc03870d975be05c67f14a30b/base/compiler/ssair/inlining.jl#L1668-L1669

test/plugin_testsetup.jl
vchuravy: See https://github.com/JuliaGPU/GPUCompiler.jl/pull/634/files#diff-1bf557ada55697d453a6ccf81e9c263e46e573e324c453c4ecffb70994f36c37R504 for context.

vchuravy: I have a double extension problem. GPUCompiler provides has a GPUInterpreter.
Both GPUCompiler/CUDA and Enzyme want to modify the rules being applied.

But when we are applying Enzyme to CUDA code we must "inherit" the rules from CUDA, up to now Enzyme had a EnzymeInterpreter, but I would like to get rid of that.

But Enzyme rules shouldn't apply to CUDA code by default. However I also need to teach in an extensible matter GPUCompiler about `autodiff` such that:


```julia
function kernel(args...)
    autodiff(f, ....)
end
@cuda kernel(args...)
```

works.

test/plugin_testsetup.jl
aviatesk: Ah, I just realized that overloading `src_inlining_policy` wouldn't be enough.. We need to overload `retrieve_ir_for_inlining` too, but it doesn't take `info::CallInfo`, so maybe we need to tweak the interface..

But I believe this approach (overloading `inlining_policy`) works at least for pre-1.11.

> Haha, then I have misunderstood the comment in:
> 
> https://github.com/JuliaLang/julia/blob/b9d9b69165493f6fc03870d975be05c67f14a30b/base/compiler/ssair/inlining.jl#L1668-L1669

It seems like I’ve ended up betraying my past self.

vchuravy: Yeah I think you once told me to extend handle_call

test/plugin_testsetup.jl
wsmoses: I think presently Enzyme does more than that as well. To rough approximation, it does the following as its entire compilation step

1) "Before anything else happens"
   Set each llvmf to know about its worldage, methodinstance, and return type: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L6703
   Make things inline ready [e.g. remove some tbaa which is broken] https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L6705
   Rewrite some nvvm and related intrinsics https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L6709
   mark various type unstable calls as inactive and change inttoptr'd ccalls into calls by name [storing the actual int value to later restore]: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L6714
   Replace unhandled blas calls with fallback: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L6755
   Annotate types and activities: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L6894
   Mark custom rules and related as noinline: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L7035
   Lower calling convention of functino being differentiated: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L7465
   
2) Optimization pipeline: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L7514
    Currently we use a modified optimization pipeline that also adds new passes which we found to be critical for performance (namely the new jl_inst_simpliy pass among others for interprocedural dead arg elim)
    
3) AD 
    First we run a julia analysis pass if the fn differentiated was a closure and requested we error if it is written https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L7659
    Upgrading some memcpy's to load/store: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L7799
    Actually generating the derivatives: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L7801
     Inverse of the preserve nvvm pass above: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L7897
     Restoring the actual inttoptr => function name from above https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L8074
     Other immediate post Enzyme passes: https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L8085 and https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L8087 and https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L8105 and https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L8129
     
Post Enzyme optimization (https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler.jl#L8131):
   * This includes running new passes to fix garbage collection/etc but presumably can just be scheduled

vchuravy: It would be more useful to think not about "what does Enzyme do", but "what information does Enzyme need and when".

The actual plugin PR is 
https://github.com/JuliaGPU/GPUCompiler.jl/pull/633 we can add more callbacks, as long as the callbacks only trigger if we detect that a marker function is present in the module.

The issue is that Enzyme requires an orthogonal composability axes than the job level.

vchuravy: E.g. what are the interfaces you need, also what his historic and what are design constraints.

As an example, I would like the LLVM IR to be serializable without references to runtime pointers. 
Otherwise caching will be impossible and CUDA.jl is already caching it's LLVM IR (post optimizations currently so we could just throw away the Enzyme metadata)

wsmoses: I think all of the before Enzyme stuff is necessary atm. The caveat being that if there was a better way to make sure functions had better names than us doing the converstion from an inttoptr to the name and restoring it, that would be great. So we definitely need a "right after IR is generated pass plugin"

We presently need ideally other hooks into the optimization pipeline to add the other passes we run. For example https://github.com/EnzymeAD/Enzyme.jl/blob/3c0871d54b85e72a8b0dc2e2e5d48d2eb6e0a95e/src/compiler/optimize.jl#L2522

re 3) We need ways to know the full context of the julia AD request. I think this is basically just the gpucompiler config object [and the restore inttoptr name map]

vchuravy: So a caveat of the implementation here is that we no longer create a EnzymeConfig object when inside a CUDA compilation. We still do when we do a host compilation, but wouldn't anymore for nested compilation.

We would need in the lowering from `autododiff` to llvmcall of autodiff capture the necessary information and encode that on the calling side (e.g. similar to the Clang plugin for Enzyme).

>  "right after IR is generated pass plugin"

Right, that's what I added in https://github.com/JuliaGPU/GPUCompiler.jl/pull/633#discussion_r1787747311


So keep in mind that we are talking about two compilation modes, first of all the inside CUDA.jl and the second one for nested.

For CPU nested you would still have a lot more control over code flow and could do all the blas things and inttoptr, but those are errors already on the GPU.

Issue #635, opened 2024-09-29
1.11 error AssertionError: Base.isdispatchtuple


wsmoses:
```julia
(f, tt, world, typeof(f)) = (Enzyme.Compiler.add_one_in_place, Tuple{Any}, 0x000000000000683a, typeof(Enzyme.Compiler.add_one_in_place))
  
  Stacktrace:
    [1] methodinstance
      @ ~/.julia/packages/GPUCompiler/2CW9L/src/jlgen.jl:82 [inlined]


 function nested_codegen!(mode::API.CDerivativeMode, mod::LLVM.Module, f, tt, world)
+    @show f, tt, world, typeof(f)
     funcspec = GPUCompiler.methodinstance(typeof(f), tt, world)
     nested_codegen!(mode, mod, funcspec, world)
 end
```

So essentially here we request a generic form of the function `add_one_in_place` requesting it for an any input. This works on 1.10, but fails on 1.11 hitting the aforementioned assertion error

maleadt:
```
  isdispatchtuple(T)

  Determine whether type T is a tuple "leaf type", meaning it could appear as a type signature in dispatch and has
  no subtypes (or supertypes) which could appear in a call.
```

`Tuple{Any}` is not a valid type to request code generation for. This is also not supported by Base; see the warning emitted by `code_llvm`:

```
julia> code_llvm(identity, Tuple{Any})
; WARNING: This code may not match what actually runs.
;  @ operators.jl:522 within `identity`
define nonnull {}* @julia_identity_128({}* noundef nonnull readonly %0) #0 {
top:
  ret {}* %0
}
```

wsmoses:
Worked around in a different way

wsmoses:
Okay I take it back, this remains blocking for us in custom rules in 1.11.

@gbaraldi you had mentioned some ideas working around this?

cc @ChrisRackauckas 

@maleadt even with that warning as is, can we somehow still work around this and get the equivalent of that codellvm at the bottom?

PR #644, opened 2024-10-17
Ensure methodinstance can be used on abstract types

wsmoses:
Needed for Enzyme custom rules in Julia 1.11

Let's us revert having to vendor it ourselves in https://github.com/EnzymeAD/Enzyme.jl/pull/1981/files

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -71,33 +71,7 @@ function generic_methodinstance(@nospecialize(ft::Type), @nospecialize(tt::Type)
     return mi::MethodInstance
 end
 
-# on 1.11 (JuliaLang/julia#52572, merged as part of JuliaLang/julia#52233) we can use
-# Julia's cached method lookup to simply look up method instances at run time.
-if VERSION >= v"1.11.0-DEV.1552"
-
-# XXX: version of Base.method_instance that uses a function type
-@inline function methodinstance(@nospecialize(ft::Type), @nospecialize(tt::Type),
-                                world::Integer=tls_world_age())
-    sig = signature_type_by_tt(ft, tt)
-
-    mi = ccall(:jl_method_lookup_by_tt, Any,
-               (Any, Csize_t, Any),
-               sig, world, #=method_table=# nothing)
-    mi === nothing && throw(MethodError(ft, tt, world))
-    mi = mi::MethodInstance
-
-    # `jl_method_lookup_by_tt` and `jl_method_lookup` can return a unspecialized mi
-    if !Base.isdispatchtuple(mi.specTypes)
-        mi = CC.specialize_method(mi.def, sig, mi.sparam_vals)::MethodInstance
-    end
-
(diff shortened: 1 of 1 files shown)


maleadt:
@vtjnash Can you take a look at this? Are these queries valid for non-dispatch tuples?

wsmoses:
x/ref https://github.com/JuliaGPU/GPUCompiler.jl/issues/635

src/jlgen.jl
```diff
@@ -79,7 +79,6 @@ if VERSION >= v"1.11.0-DEV.1552"
 @inline function methodinstance(@nospecialize(ft::Type), @nospecialize(tt::Type),
                                 world::Integer=tls_world_age())
     sig = signature_type_by_tt(ft, tt)
-    @assert Base.isdispatchtuple(sig)   # JuliaLang/julia#52233
```
vtjnash, line 82: The underlying function this ccalls may assert of segfault without this. There is probably a significant argument to be made that you should be using the public function `which(sig)` here, which doesn't have that footgun, but it currently doesn't know that it could optimize the `isdispatchtuple` case by calling `jl_method_lookup_by_tt`.

wsmoses:
@maleadt how is this then, we can still use the fast path when legal, but when illegal we use the same path as 1.10.

maleadt:
The whole idea of the new upstream API was to get rid of the generator, which is fundamentally broken (e.g. https://github.com/JuliaGPU/GPUCompiler.jl/issues/530), so we're not adding that back.

vchuravy:
Maybe to avoid an X then Y, problem. From the perspective of the Julia compiler and runtime system, the query is ill-formed and, as far as I can tell, Julia itself never does this query.

So the question would be "Why is Enzyme forming this query" and "Can we make this query in Enzyme well-posed"?

wsmoses:
This primarily comes up in the context of Enzyme custom rules.

Specifically we first call GPUCompiler to generated llvm for a corresponding augmented forward derivative, which returns the original result and a tape.

We then call MethodInstance to get the llvm for the corresponding reverse rule which has the same arguments, plus this additional tape arg.

suppose we were differentiating sin(x)

we would want Enzyme.augmentedforward, Active(1.0) for the forward pass and Enzyme.reverse Active(1.0), tape for the reverse pass where tape is an element of the return of the first call.

Often times (and it’s completely legal) for Julia to not be able to statically determine the concrete tape type, so this could be Any for example.

which leads us to asking for Tuple{typeof(reverse), Active{Float64}, Any}.

Of course the user could specify a concrete tape type so there’s usually a reason why it isn’t (as an example within CUDA.jl we need to compute the tape type), and in these cases performance wise it’s usually better to pass this in as an any (eg as if nospecalize was in the arg). This is the type of case causing the failure.

This worked perfectly fine with the 1.10 and before code in GPUCompiler, but now throws an assertion.

PR #653, opened 2024-12-01
Use Cached type inference

wsmoses:


vchuravy (approved):
Since we already have the `mi` we can use it directly instead of deconstructing it.

wsmoses:
for better or worse sadly it takes ~1 hour [52 mins or so] atm. From our tests it looks to be spending most of the time on inference, which we're hoping to bring back down.

PR #711, opened 2025-08-18
Fix deferred_codegen registration

vchuravy:
According to @gbaraldi this was running into issues on 1.11 when code-coverage was enable (since Base dropped the registration of the ccallable)
and on 1.12 this is going to be broken by https://github.com/JuliaLang/julia/pull/56987

### DIFF
--- a/src/driver.jl
+++ b/src/driver.jl
@@ -129,7 +129,7 @@ const deferred_codegen_jobs = Dict{Int, Any}()
 
 # We make this function explicitly callable so that we can drive OrcJIT's
 # lazy compilation from, while also enabling recursive compilation.
-# Julia 1.11 and co broke @ccallable so we have to do this manually in __init__
+# see `register_deferred_codegen`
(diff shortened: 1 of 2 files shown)


src/driver.jl
```diff
@@ -149,6 +150,33 @@ end
...
+            LLVM.API.LLVMOrcCSymbolMapPair(name, symbol)
+        else
+            LLVM.API.LLVMJITCSymbolMapPair(name, symbol)
+        end
+
+        mu = LLVM.absolute_symbols(Ref(map))
+        LLVM.define(jd, mu)
+        addr = lookup(jljit, "deferred_codegen")
+        @assert addr != C_NULL "Failed to register deferred_codegen"
+    end
```
gbaraldi, line 176: Should we ccall it here just to check if it's working? I guess the address is enough but not sure

gbaraldi:
This isn't pretty, but it's also not very different from what was there before.

PR #718, opened 2025-09-09
add_input_arguments! for other backends

simeonschaub:
Allows other backends to pass additional hidden arguments that can be accessed through intrinsics. Required for OpenCL device-side RNG support, where additional shared memory must be passed as arguments to the kernel.

Replaces https://github.com/JuliaGPU/GPUCompiler.jl/pull/717


src/irgen.jl
```diff
@@ -923,3 +926,160 @@ function kernel_state_to_reference!(@nospecialize(job::CompilerJob), mod::LLVM.M
...
+        fn = LLVM.name(f)
+        ft = function_type(f)
+        LLVM.name!(f, fn * ".orig")
+        # create a new function
+        new_param_types = LLVMType[parameters(ft)...]
+
+        for intr_fn in used_intrinsics
+            llvm_typ = convert(LLVMType, kernel_intrinsics[intr_fn].typ)
+            push!(new_param_types, llvm_typ)
+        end
```
vchuravy, line 977: The goal is to reuse the function for the kernel state rewrite? One issue is that for Metal we are adding things at the end following it's ABI,

Whereas the kernel argument would be the first argument. Also I recall Metal having a whole kerfuffle around requiring metadata to describe arguments?

PR #725, opened 2025-10-04
change call to `codeinfo_for_const` to work on newer nightlies

KristofferC:


KristofferC:
There are some changes in the stdlib/Test that requires changes here, 🤖 says:

  - **`push_testset`/`pop_testset` removed** (Julia 1.13.0-DEV.1044)
    - Commit: [bb36851288](https://github.com/JuliaLang/julia/commit/bb36851288) - "use ScopedValues for TestSets"
    - These functions were replaced with ScopedValues-based approach,
  but weren't actually needed since `record()` takes testset as
  explicit parameter

  - **`Test.Error` constructor signature changed** (Julia
  1.13.0-DEV.769)
    - Commit: [76d5b14c9c](https://github.com/JuliaLang/julia/commit/76d5b14c9c) - "Test: show context when a let testset errors"
    - Added `context::Union{Nothing, String}` parameter
    - Old: `Error(test_type, orig_expr, value, bt, source)`
    - New: `Error(test_type, orig_expr, value, bt, source,
  context=nothing)`

  - **`anynonpass` field type changed from `Bool` to `UInt8`** (Julia
  1.13.0-DEV.1037)
    - Commit: [364ecb3a11](https://github.com/JuliaLang/julia/commit/364ecb3a11) - "Test: Make test recording thread safe"
    - Changed to atomic UInt8 with values: `0x0=unknown`, `0x1=all passed`, `0x2=some failed`
    - Need to check `!= 0x1` instead of using `!` operator

KristofferC:
For the Test problems, I guess either moving to using `Base.runtests` directly is a good idea or alternatively just update from there,

KristofferC:
Yeah, that's why I said

> (although I am not sure that is possible right now without some tweaks to it)

LinearAlgebra uses it from its own repo: https://github.com/JuliaLang/LinearAlgebra.jl/blob/7b21cabb1e81b719e93ab612937c0ded809fbe6e/.ci/run_tests.jl#L8 and that barely works because julia considers it an stdlib. Some follow-up to https://github.com/JuliaLang/julia/pull/56558 should make it possible to use it for packages as well.

kshyatt:
I'm making a PR to do that now

PR #737, opened 2025-11-10
unwrap ExceptionStack in code_ handling

vchuravy:
We inform users to use:

```
Catch this exception as `err` and call `code_typed(err; interactive = true)`
```

But in the REPL this is often not useful advice since the exception of interest is actually wrapped...
This biggest issue with this PR is that we are technically committing type-piracy, but this improves the UX for working with compiler errors from GPUCompiler a lot.

I think there is a second error type that we can get which is when the error happens in a file that is being `include`'d

### DIFF
--- a/src/error.jl
+++ b/src/error.jl
@@ -25,6 +25,19 @@ end
 Base.show(io::IO, err::KernelError) = showerror(io, err)
 
 
+"""
+    unwrap_error(err)
+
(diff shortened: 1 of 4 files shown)


maleadt:
> `` Catch this exception as `err` and call `code_typed(err; interactive = true)` ``

I never was a big fan of this; `@device_code_typed` is always the tool I use. But for large applications I guess it may be useful.

PR #741, opened 2025-11-14
use `CC.MAX_INLINE_COST` instead of `typemax(Int)`

simeonschaub:
as cost threshold for `always_inline` to avoid integer overflow

fix JuliaGPU/AMDGPU.jl#846

alternative to JuliaLang/julia#60121


aviatesk:
Yeah, I think we need something like https://github.com/JuliaLang/julia/pull/48257 as a complete fix.

PR #743, opened 2025-11-23
Imaging-mode PLT codegen and relocatable TLS

joelreymont:
- Enable imaging-mode PLT lookups only on Julia versions that expose use_jlplt,
    preserving older compat.
  - Make StaticCompiler host targets emit relocatable TLS getters by rewriting
    intrinsic and inttoptr forms and ensuring the pass runs after upstream
    transforms.
  - Update docs and fix the native ccall test message match.
  - Tests: julia --project -q -e 'using Pkg; Pkg.test()'.

These changes are required for runtime-linked StaticCompiler targets. Without these GPUCompiler patches, executables that link the Julia runtime would bake absolute TLS addresses and crash at startup.

### DIFF
--- a/src/optim.jl
+++ b/src/optim.jl
@@ -1,12 +1,12 @@
 # LLVM IR optimization
 
-function optimize!(@nospecialize(job::CompilerJob), mod::LLVM.Module; opt_level=2)
+function optimize!(@nospecialize(job::CompilerJob), mod::LLVM.Module; opt_level = 2)
     tm = llvm_machine(job.config.target)
 
     global current_job
     current_job = job
 
-    @dispose pb=NewPMPassBuilder() begin
+    @dispose pb = NewPMPassBuilder() begin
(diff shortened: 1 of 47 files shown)


joelreymont:
Yes, that Julia PR requires this one.

src/optim.jl
```diff
@@ -19,6 +19,14 @@ function optimize!(@nospecialize(job::CompilerJob), mod::LLVM.Module; opt_level=
         end
         run!(pb, mod, tm)
     end
+
+    # Make sure any lingering TLS getters are rewritten even if upstream LLVM passes
+    # transformed them before the GPULowerPTLSPass had a chance to run.
+    if occursin("StaticCompilerTarget", string(typeof(job.config.target))) &&
```
vchuravy: This is the opposite of the code pattern we use, we should never hard code a specific target here, but rather expose hooks for a target to configure the pipeline

PR #746, opened 2025-11-26
Enable users to map from GV to Julia value

vchuravy:
Two fold motivation.

Currently, we perform the JIT favourite trick on all globals, we inline the runtime pointers, this makes the resulting code non-relocatable. A concrete example of that would be something like:

```
function f(x)
    if x === :x
    end
end
```

The memory location of `:x` changes upon session restart (as do other things like Types, Singletons), and for objects like this Julia performs a simple pointer comparison. So we can actually support this operation on the GPU.

Yet, we cannot cache the resulting kernel, or if we do, we will perform the wrong computation. So what we must do instead is to initialize the global after loading it from cache. 

Now that is a tricky affair (backend dependent), we can do it on LLVM IR, I believe AMDGPU could also do it, the CPU backends can use something like `JITDylib`, but I don't think we can do it on the obj file from Nvidia, and I don't about other backends.

So we should likely mark compilations that use globals with runtime pointers as ineligible for caching.

The second motivation stems from Enzyme where we would like to have a mapping from global to Julia object, and it can be hard to figure out which globals are Julia objects and which ones are not.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -253,12 +253,12 @@ end
             if !ondisk_hit && path !== nothing && disk_cache_enabled()
                 @debug "Writing out on-disk cache" job path
                 mkpath(dirname(path))
-                if haskey(asm.meta, :gv_to_value)
+                if haskey(asm[2].meta, :gv_to_value)
                     # TODO: Serializer cannot handle Core.IntrinsicFunction
                     # We kinda want Julia to serialize the values in `gv_to_value` in the pkgimg and us just having to store an index
                     # for now we just empty them out
                     # We would need to remove the initializers from LLVM IR as well to be correct, and then link these in at runtime
-                    empty!(asm.meta.gv_to_value)
+                    empty!(asm[2].meta.gv_to_value)
                 end
                 entry = DiskCacheEntry(src.specTypes, cfg, asm)
 
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -253,6 +253,13 @@ end
             if !ondisk_hit && path !== nothing && disk_cache_enabled()
                 @debug "Writing out on-disk cache" job path
                 mkpath(dirname(path))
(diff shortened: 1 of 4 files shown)


src/execution.jl
```diff
@@ -253,6 +253,13 @@ end
             if !ondisk_hit && path !== nothing && disk_cache_enabled()
                 @debug "Writing out on-disk cache" job path
                 mkpath(dirname(path))
+                if haskey(asm[2], :gv_to_value)
+                    # TODO: Serializer cannot handle Core.IntrinsicFunction
+                    # We kinda want Julia to serialize the values in `gv_to_value` in the pkgimg and us just having to store an index
+                    # for now we just empty them out
+                    # We would need to remove the initializers from LLVM IR as well to be correct, and then link these in at runtime
+                    empty!(asm[2].gv_to_value)
```
vchuravy: We should likely just give up on caching this for now if `!isempty(gv_to_value)`

For proper caching, we would want to create an `IdDict` in the current parent module  and instead of saving values we would want to save "offsets/keys" into that `IdDict` so that Julia performs the pkgimg reolcation for us.

src/jlgen.jl
```diff
@@ -786,25 +808,73 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
...
+            if ptr in gvalues
+                gv_to_value[LLVM.name(gv)] = ptr
+            end
+            LLVM.initializer!(gv, nothing)
         end
+        @assert length(gv_to_value) == length(gvalues)
     end
+    # It's valid to call Base.unsafe_pointer_to_objref on values(gv_to_value),
+    # but we may not be able to "easily" obtain the pointer back later.
+    # (Types, etc, disallow Base.pointer_from_objref on them.)
```
maleadt: What does this comment apply to?

vchuravy:
Stale

Issue #611, opened 2024-08-05
JIT session error triggered by precompilation


pxl-th:
Using PrecompileTools.jl to [precompile](https://github.com/JuliaNeuralGraphics/GaussianSplatting.jl/blob/main/ext/GaussianSplattingCUDAExt/GaussianSplattingCUDAExt.jl) GaussianSpaltting.jl triggers following error on 1.10 when executing code after precompilation:

```julia
JIT session error: Symbols not found: [ julia.gpu.state_getter ]
JIT session error: Failed to materialize symbols: { (JuliaOJIT, { julia_#throw_inexacterror_2603 }) }
JIT session error: Failed to materialize symbols: { (JuliaOJIT, { julia_#GaussianRasterizer#16_2598 }) }
```

MWE:

The error seems to be triggered at the moment of calling `GaussianRasterizer` ctor.

```julia
using CUDA
using cuDNN
using GaussianSplatting

function main()
    kab = GaussianSplatting.gpu_backend()

    points = CuArray(rand(Float32, 3, 128))
    colors = CuArray(rand(Float32, 3, 128))
    scales = CuArray(rand(Float32, 3, 128))

    camera = GaussianSplatting.Camera(; fx=100f0, fy=100f0, width=256, height=256)
    gaussians = GaussianSplatting.GaussianModel(points, colors, scales; max_sh_degree=0)
    rasterizer = GaussianSplatting.GaussianRasterizer(kab, camera; auxiliary=false)
    return
end
main()
```

maleadt:
cc @vchuravy

KSepetanc:
PackageCompiler fails to compile CUDA with `compile=all`. As suggested in [Discourse](https://discourse.julialang.org/t/packagecompiler-fails-to-create-app-for-madnlpgpu-examodels-cudss-linear-solver/125998), this issue might be related or the root of it.

KSepetanc:
I wish to raise the importance of this issue as it blocks compilation of any CUDA.jl related project (using both PrecompileTools and PackageCompiler).

maleadt:
MWE:

```julia
module MWE

using GPUCompiler, PrecompileTools

struct CompilerParams <: AbstractCompilerParams end

@inline threadfence_block() = ccall("llvm.nvvm.membar.cta", llvmcall, Cvoid, ())
kernel() = threadfence_block()

@setup_workload begin
    source = methodinstance(typeof(kernel), Tuple{})
[...]

PR #749, opened 2025-12-22
Enable Ahead-of-Time Compilation by hiding the runtime functions in the `GLOBAL_METHOD_TABLE`

apozharski:
As discussed in https://github.com/JuliaGPU/CUDA.jl/pull/2998 and https://github.com/JuliaGPU/GPUCompiler.jl/issues/611 currently `GPUCompiler.jl` leaks nonexistant `gpu_*` llvm functions into the cpu cache making ahead of time compilation impossible for any package that uses it.

I am currently fixing this by moving these runtime methods into the method table defined in the GPUCompiler module and having the CPU versions throw errors as is done in `CUDA.jl`. This feels like somewhat of a hack, however, it seems to work and without a better understanding of what this might break it seems to be the simplest solution.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -238,3 +238,6 @@ end
         return inits
     end
 end
+## method overrides
+
+Base.Experimental.@MethodTable(GLOBAL_METHOD_TABLE)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -242,13 +242,18 @@ end
 
 Base.Experimental.@MethodTable(GLOBAL_METHOD_TABLE)
 using ExprTools: splitdef, combinedef
-macro device_function(ex)
+macro device_function(rt, ex)
     ex = macroexpand(__module__, ex)
     def = splitdef(ex)
 
-    # generate a function that errors
+    # generate a function that warns and returns the expected type
+    # FIXME: The type may not have a default constructor, what do we do then?
+    #        Currently we are using the constructor with an Int64(1) as an argument.
+    # NOTE: using Int64(1) is a bit odd. This is because Ptr(Int64(0)) == C_NULL, and julia code lowering
+    #       seems to get rid of this automatically.
     def[:body] = quote
-        error("This function is not intended for use on the CPU")
+        @warn "This function is not intended for use on the CPU something may have gone wrong"
+        $rt(1)
     end
 
     esc(quote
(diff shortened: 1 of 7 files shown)


michel2323 (approved):
CUDA instructions still leak to CPU at the LLVM level, but this unblocks the immediate AOT use case.

christiangnrd:
I'm not familiar enough with this part of Julia/GPUCompiler to meaningfully review

src/utils.jl
```diff
@@ -238,3 +238,29 @@ end
...
+    #       seems to get rid of this automatically.
+    def[:body] = quote
+        $rt(1)
+    end
+
+    esc(quote
+        $(combinedef(def))
+
+        # NOTE: no use of `@consistent_overlay` here because the regular function errors
+        Base.Experimental.@overlay($(GPUCompiler).GLOBAL_METHOD_TABLE, $ex)
```
maleadt, line 262: This is wrong, or at least breaking. Not all back-ends use the GLOBAL_METHOD_TABLE, e.g., CUDA.jl has its own.

apozharski: Would a solution here be to support passing a method table to this macro and only default to `GLOBAL_METHOD_TABLE`? I suspect that this would still be breaking but if that is acceptable, I would be happy to do make the necessary changes in `CUDA.jl` etc to support this.

maleadt: But the macro is invoked from within GPUCompiler.jl's `runtime.jl`, so how could it ever point to the CUDA.jl method table?

maleadt: Maybe we ought to make it mandatory for back-ends to use a stacked overlay table that includes the GPUCompiler one. Thoughts, @vchuravy?

apozharski: Ahhh wait, this was to avoid the `compile` calls here in the runtime module from leaking, I forgot... But I am a bit confused how this was working regardless (at least on my end I haven't seen any regressions from this). 

Then yea, I don't have a good workaround but can confirm that this revert broke AOT compilation again.

maleadt:
@apozharski @KSepetanc Please try with https://github.com/JuliaGPU/GPUCompiler.jl/pull/808

KSepetanc:
No compile issues on 1.12 with 808.

Will try it on Julia 1.13 and report once MadNLPGPU related issues are resolved so CUDA v6 can be used.

apozharski:
> @apozharski @KSepetanc Please try with #808

MadNLP was still reeling from CUDA v6 updates so I will check `libMad` once things are working on my end. I suspect the weak linkage in #808 will work with no issues.

PR #750, opened 2026-01-07
Allow semi-concrete interpretation

maleadt:
Try out https://github.com/JuliaGPU/GPUCompiler.jl/issues/382


maleadt:
@aviatesk Seems to work fine :-)

PR #752, opened 2026-01-14
Enable users to map from GV to Julia value and initialize them whenever possible

wsmoses:
Simplified version of https://github.com/JuliaGPU/GPUCompiler.jl/pull/746 that _just_ adds the additional argument with remaining a NFC for other initialization/IR changes.

Context: it became somewhat more urgent to get this in for Enzyme/Reactant to support 1.12 + relocation/precompilation. Rather than do homebrew solutions, having this would enable a significant simplification.

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -652,62 +652,6 @@ end
     CompilationPolicyExtern = 1
 end
 
-@static if VERSION < v"1.13.0-DEV.623"
-    import Libdl
-
-    const HAS_LLVM_GVS_GLOBALS = Libdl.dlsym(
-        unsafe_load(cglobal(:jl_libjulia_handle, Ptr{Cvoid})), :jl_get_llvm_gvs_globals, throw_error=false) !== nothing
-
-    const AL_N_INLINE = 29
-
-    # Mirrors arraylist_t
-    mutable struct ArrayList
-        len::Csize_t
-        max::Csize_t
-        items::Ptr{Ptr{Cvoid}}
-        _space::NTuple{AL_N_INLINE, Ptr{Cvoid}}
-
-        function ArrayList()
-            list = new(0, AL_N_INLINE, Ptr{Ptr{Cvoid}}(C_NULL), ntuple(_ -> Ptr{Cvoid}(C_NULL), AL_N_INLINE))
-            list.items = Base.pointer_from_objref(list) + fieldoffset(typeof(list), 4)
-
-            finalizer(list) do list
-                if list.items != Base.pointer_from_objref(list) + fieldoffset(typeof(list), 4)
-                    Libc.free(list.items)
-                end
-            end
-            return list
-        end
-    end
-
-    function get_llvm_global_vars(native_code::Ptr{Cvoid})
-        gvs_list = ArrayList()
-        GC.@preserve gvs_list begin
-            p_gvs = Base.pointer_from_objref(gvs_list)
-            @ccall jl_get_llvm_gvs_globals(native_code::Ptr{Cvoid}, p_gvs::Ptr{Cvoid})::Nothing
-            gvs = Vector{Ptr{LLVM.API.LLVMOpaqueValue}}(undef, gvs_list.len)
-            items = Base.unsafe_convert(Ptr{Ptr{LLVM.API.LLVMOpaqueValue}}, gvs_list.items)
-            for i in 1:gvs_list.len
-                gvs[i] = unsafe_load(items, i)
-            end
-        end
-        return gvs
-    end
-
-    function get_llvm_global_inits(native_code::Ptr{Cvoid})
-        inits_list = ArrayList()
-        GC.@preserve inits_list begin
-            p_inits = Base.pointer_from_objref(inits_list)
-            @ccall jl_get_llvm_gvs(native_code::Ptr{Cvoid}, p_inits::Ptr{Cvoid})::Nothing
-            inits = Vector{Ptr{Cvoid}}(undef, inits_list.len)
-            for i in 1:inits_list.len
-                inits[i] = unsafe_load(inits_list.items, i)
-            end
-        end
-        return inits
-    end
-end
-
 """
     precompile(job::CompilerJob)
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -182,3 +182,59 @@ function kernels(mod::LLVM.Module)
     end
     return vals
 end
+
+@static if VERSION < v"1.13.0-DEV.623"
+    import Libdl
+
+    const HAS_LLVM_GVS_GLOBALS = Libdl.dlsym(
+        unsafe_load(cglobal(:jl_libjulia_handle, Ptr{Cvoid})), :jl_get_llvm_gvs_globals, throw_error=false) !== nothing
+
+    const AL_N_INLINE = 29
+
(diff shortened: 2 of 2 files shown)


src/jlgen.jl
```diff
@@ -766,25 +791,80 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
...
+                native_code::Ptr{Cvoid}, num_gvars::Ptr{Csize_t},
+                C_NULL::Ptr{Cvoid}
+            )::Nothing
+            gvalues = Vector{Ptr{Cvoid}}(undef, num_gvars[])
+            @ccall jl_get_llvm_gvs(
+                native_code::Ptr{Cvoid}, num_gvars::Ptr{Csize_t},
+                gvalues::Ptr{Cvoid}
+            )::Nothing
+        else
+            # Sigh on older version of Julia we have to use `arraylist_t` which doesn't have a Julia API.
```
maleadt:
```suggestion
            # On older version of Julia we have to use `arraylist_t` which doesn't have a Julia API.
```

src/jlgen.jl
```diff
@@ -766,25 +791,80 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
...
+            ptr = reinterpret(Ptr{Cvoid}, convert(UInt, init))
+            if ptr in gvalues
+                gv_to_value[LLVM.name(gv)] = ptr
+            end
+        end
+        @assert length(gv_to_value) == length(gvalues)
     end
+    # It's valid to call Base.unsafe_pointer_to_objref on values(gv_to_value),
+    # but we may not be able to "easily" obtain the pointer back later.
+    # (Types, etc, disallow Base.pointer_from_objref on them.)
```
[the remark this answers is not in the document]
wsmoses: it applies to the values whose pointer is stored in in gv_to_value. Specifically if you have a global representing 2.0, we have gvalue = pointer_from_objref(2.0). There's an issue however that performing pointer_from_objref on that type is invalid [since its not mutable], hence we store the actual pointer

src/jlgen.jl
```diff
@@ -852,8 +852,14 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
...
         end
-        gvalues = Set(gvalues)
+	idx = 1
         for gv in globals(llvm_mod)
+	    if !haskey(metadata(gv), "julia.constgv")
+	       continue
+	    end
+	    gv_val = gvalues[idx]
+	    idx += 1
+	    gv_to_value[LLVM.name(gv)] = gv_val
```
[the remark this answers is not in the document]
wsmoses: I can only get it to happen during precompilation, and I'm really unsure how to write a test in precompilation

test/native.jl
```diff
@@ -36,23 +36,34 @@ end
...
         mod = @eval module $(gensym())
             @noinline inner(x) = x+1
-            function outer(x)
+        function outer(x, sym)
+            if sym == :a
                 return inner(x)
             end
+            return x
+            end
         end
```
vchuravy: The formatter broke formatting here

src/jlgen.jl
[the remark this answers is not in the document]
vchuravy: Can you explain the need for the PRs? What doesn't work?

Julia has a very strict "We don't backport features" policy and opening three PRs with no explanation is not that great to convince folks that something is needed.

test/native/precompile.jl
```diff
@@ -14,11 +14,23 @@ precompile_test_harness("Inference caching") do load_path
             return
         end
 
+        function kernel_w_global(A, x, sym)
+            if sym == :A
+                A[1] = x
+            end
+            return
+        end
+
```
vchuravy: @wsmoses please add an example here of where one would get a null initialized GV.

src/jlgen.jl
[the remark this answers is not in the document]
oscardssmith: @wsmoses can you at least tag the PRs as backports? (and what they're backports of?)

test/native/precompile.jl
wsmoses: okay I genuinely how no idea how to reproduce in the test harness, all my attempts have failed

src/jlgen.jl
```diff
@@ -766,24 +795,89 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
...
+            if init isa LLVM.ConstantExpr && opcode(init) == LLVM.API.LLVMIntToPtr
+                init = operands(init)[1]
+            end
+            if !(init isa LLVM.ConstantInt)
+                @show init
+            end
+            ptr = reinterpret(Ptr{Cvoid}, convert(UInt, init))
+            if ptr in gvalues
+                gv_to_value[LLVM.name(gv)] = ptr
+            end
```
vchuravy: This seems just wrong?

gbaraldi:
The check should be something like this
```julia
[5:28 PM]b = unsafe_load(Ptr{Ptr{Cvoid}}(cglobal(:jl_libjulia_handle)))
Libdl.dlsym(b, :jl_get_llvm_gv_inits, throw_error=false)
```

src/driver.jl
```diff
@@ -256,6 +256,7 @@ const __llvm_initialized = Ref(false)
                     dyn_ir, dyn_meta = codegen(:llvm, CompilerJob(dyn_job; config))
                     dyn_entry_fn = LLVM.name(dyn_meta.entry)
                     merge!(compiled, dyn_meta.compiled)
+                    merge!(gv_to_value, dyn_meta.gv_to_value)
```
vchuravy:
```suggestion
                    if haskey(dyn_meta, :gv_to_value)
                        merge!(gv_to_value, dyn_meta.gv_to_value)
                    end    
```

src/jlgen.jl
```diff
@@ -766,23 +824,76 @@ function compile_method_instance(@nospecialize(job::CompilerJob))
...
+    else
+        @assert inits !== nothing
         for (gv_ref, init) in zip(gvs, inits)
             gv = GlobalVariable(gv_ref)
-            val = const_inttoptr(ConstantInt(Int64(init)), LLVM.PointerType())
-            initializer!(gv, val)
+            gv_to_value[LLVM.name(gv)] = init
+            # set the initializer
+            # TODO: To enable full relocation we should actually strip out the initializers here.
+            if initializer(gv) isa LLVM.PointerNull
```
vchuravy:
```suggestion
            # TODO(vc): To enable full relocation we should actually strip out the initializers here.
            if LLVM.isnull(initializer(gv))
```

maleadt (commented):
Superficially LGTM.

vchuravy:
> Should we always emit this with imaging mode for consistency? Or do we only care when we're actually precompiling

Perhaps, but that would be to disruptive. It would immediately break v1.10, and v1.11, 1.12 is a weird intermediate state, and 1.13 already uses `imaging mode`. 

So for now, this is mostly defensive against use during precompilation, and will pave the way to actually emit relocatable caches down the road.

wsmoses:
from your earlier log: https://buildkite.com/julialang/gpucompiler-dot-jl/builds/2129/steps/canvas

but it seems you might've fixed it already?

wsmoses:
since this seems approved by all, I'm going to go ahead and merge, but let you all do the release [presumably a minor version bump?]

Issue #754, opened 2026-01-16
Julia 1.12: Mismatched codeinfo hits assertion error.


wsmoses:
Specifically we hit the following error (which occurs only on 1.12, not 1.10 or 1.11):
```
ERROR: LoadError: AssertionError: Static compilation failed
Stacktrace:
  [1] compile_method_instance(job::GPUCompiler.CompilerJob)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/j4HFa/src/jlgen.jl:848
  [2] irgen(job::GPUCompiler.CompilerJob)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/j4HFa/src/irgen.jl:4
  [3] emit_llvm(job::GPUCompiler.CompilerJob; kwargs::@Kwargs{})
    @ GPUCompiler ~/.julia/packages/GPUCompiler/j4HFa/src/driver.jl:200
  [4] emit_llvm
    @ ~/.julia/packages/GPUCompiler/j4HFa/src/driver.jl:182 [inlined]
```

After deeper investigation what is happening is as follows. Our interpreter can dispatch certain methods to the native interpreter. As a consequence the methodinstance is not sufficient to correctly identify which method something is.

Within the requisite runtimecall used by gpucompiler here to find the llvm info:
```
79	extern "C" JL_DLLEXPORT_CODEGEN
80	void jl_get_function_id_impl(void *native_code, jl_code_instance_t *codeinst,
81	        int32_t *func_idx, int32_t *specfunc_idx)
82	{
83	    jl_native_code_desc_t *data = (jl_native_code_desc_t*)native_code;
84	    if (data) {
(rr) 
85	        // get the function index in the fvar lookup table
86	        auto it = data->jl_fvar_map.find(codeinst);
87	        if (it != data->jl_fvar_map.end()) {
88	            std::tie(*func_idx, *specfunc_idx) = it->second;
89	        }
90	    }
91	}
92	
```

the codeinstance provided is not found.

The corresopnding codeinstance we had was:
```
(rr) p codeinst
$15 = (jl_code_instance_t *) 0x7672abb5ce20
(rr) p jl_(codeinst->def)
(::Type{ArgumentError})(String) from (::Type{ArgumentError})(AbstractString)
```

However the contents of jl_fvar_map were:
```
(rr) p jl_((data->jl_fvar_map.begin())->first.def)
[...]

PR #755, opened 2026-01-16
Fix codeinstance mapping when using nested interpretation

wsmoses:
see https://github.com/JuliaGPU/GPUCompiler.jl/issues/754


maleadt (approved):

wsmoses:
No we need to retain the MI for each llvm function in order to determine where to apply custom rules, what types correspond to which LLVM args/instructions, among other things.

PR #762, opened 2026-01-27
[Metal] `sincos` parameter attributes

### LINKED ISSUE #663: [Metal] Add sincos intrinsic LLVM annotations
> declare half @air.sincos.f16(half, half* nocapture writeonly)
> 
> vs
> 
> declare half @air.sincos.f16(half, i64)
> 
> So the problem is we aren't passing a pointer, but an `i64`. This can be fixed by using `@typed_ccall` from LLVM.jl instead of plain `ccall`.
> 
> It's also probably useful to add those annotations to https://github.com/JuliaGPU/GPUCompiler.jl/blob/47a4a8b5f6f6737628b2307e76203da262f164a6/src/metal.jl#L1161-L1203
> 
> 
>  

 _Originally posted by @maleadt in [#530](https://github.com/JuliaGPU/Metal.jl/issues/530#issuecomment-2633230082)_

christiangnrd:
Close #663. This does NOT resolve issue #761 but I wrote it while troubleshooting it and figured I may as well open a PR.

### DIFF
--- a/src/metal.jl
+++ b/src/metal.jl
@@ -1016,9 +1016,25 @@ function annotate_air_intrinsics!(@nospecialize(job::CompilerJob), mod::LLVM.Mod
             changed = true
         end
 
+        function add_param_attributes(idx, names...)
+            param_attrs = parameter_attributes(f, idx)
+            for name in names
+                push!(param_attrs, EnumAttribute(name, 0))
+        # sincos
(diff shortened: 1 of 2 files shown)


maleadt (approved):
LGTM, thanks.

It would be good to start adding tests for things like this though. Should be reasonable with the FileCheck infrastructure we have nowadays.

PR #782, opened 2026-04-22
Metal: Configure TTI so that InferAddressSpaces works.

### LINKED ISSUE #781: Metal kernel reads zeros from constant globals with dynamic indexing
Metal kernels silently read `0.0` from LLVM constant globals that contain aggregate data (tuples, `SMatrix`, arrays) when accessed with dynamic indices. Scalar constants and statically-indexed aggregates work correctly.

## Reproducer

```julia
using Metal

function kernel_broken(out)
    i = Metal.thread_position_in_grid_1d()
    t = (1.0f0, 2.0f0, 3.0f0, 4.0f0)
    @inbounds out[i] = t[i]
    return nothing
end

out = MtlArray(ones(Float32, 4))
@metal threads=4 kernel_broken(out)
Array(out)  # [0.0, 0.0, 0.0, 0.0]  <-- should be [1.0, 2.0, 3.0, 4.0]
```

Replacing the dynamic index `t[i]` with a static index `t[2]` produces correct results. Passing the tuple as a kernel argument also works.

## Environment

- Metal.jl: v1.9.3
- GPUCompiler: v1.9.1
- Julia 1.12.5 (LLVM 18.1.7)
- macOS (Apple M2)

I have a detailed IR-level analysis of the root cause and a working local fix in GPUCompiler. Happy to share if useful. However, if a fix is more appropriate elsewhere, please let me know.

maleadt:
Fixes https://github.com/JuliaGPU/GPUCompiler.jl/issues/781

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,7 +23,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 [compat]
 ExprTools = "0.1"
 InteractiveUtils = "1"
-LLVM = "9.3"
+LLVM = "9.5"
 Libdl = "1"
 Logging = "1"
 PrecompileTools = "1"
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -37,6 +37,10 @@ end
 
 llvm_datalayout(target::AbstractCompilerTarget) = DataLayout(llvm_machine(target))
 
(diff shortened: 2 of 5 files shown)


haakon-e (commented):
Nice. I can't speak to the contents of `src/metal.jl`, but `test/metal.jl` does seem to test the issue I faced. Thanks!

PR #785, opened 2026-04-24
Improve the pipeline with callbacks

maleadt:
Uses https://github.com/JuliaLLVM/LLVM.jl/pull/524 to improve our pipeline, adding the callbacks needed to e.g. avoid manual NVVMReflect. Also changes the NVVMReflect semantics to match LLVM's, which will need some adaptation in CUDA.jl (though nothing crucial warranting a breaking release). Finally, also align the pipeline with Julia's one more closely.


wsmoses:
cc @gbaraldi @giordano 

this added an aggressiveinstcombine pass, outside of the instcombine gate, can this also be gated?

maleadt:
Sure.

PR #788, opened 2026-05-05
Revert "Enable Ahead-of-Time Compilation by hiding the runtime functions in the `GLOBAL_METHOD_TABLE`"

maleadt:
Reverts JuliaGPU/GPUCompiler.jl#749. The PR puts things in the global method table, which isn't used by back-ends that define their own.


KSepetanc:
Any other fix for AOT compile?

PR #791, opened 2026-05-06
Turn instcombine into a flag (and use instsimplify otherwise)

wsmoses:
cc @gbaraldi @vchuravy @giordano

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -226,6 +226,9 @@ end
 # Has the runtime available and does not require special handling
 uses_julia_runtime(@nospecialize(job::CompilerJob)) = false
 
+# whether to use instcombine or instsimplify
+use_instcombine(@nospecialize(job::CompilerJob)) = true
+
 # Is it legal to run vectorization passes on this target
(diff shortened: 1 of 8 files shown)


src/interface.jl
```diff
@@ -226,6 +226,9 @@ end
 # Has the runtime available and does not require special handling
 uses_julia_runtime(@nospecialize(job::CompilerJob)) = false
 
+# whether to use instcombine or instsimplify
+use_instcombine(@nospecialize(job::CompilerJob)) = true
```
maleadt: I want to avoid an ad-hoc interface for such a specific thing. I was going to think about some options struct like used by LLVM(.jl), but haven't done anything concretely yet.

wsmoses: Do you have a suggestion we can use for the moment?

This is an internal api regardless so should be safe to change

maleadt: Internal or not, the fact that it's a method makes it part of the interface and thus breaking if we change it.
I'll vibe up something.

wsmoses:
@maleadt is this good to go?

maleadt (approved):

PR #793, opened 2026-05-09
Consider assertions for runtime module

vchuravy:
While testing https://github.com/JuliaLang/julia/pull/61749 on GPUCompiler, I noticed that the runtime module would lack the flags set by Julia with assertion mode.

### DIFF
--- a/src/rtlib.jl
+++ b/src/rtlib.jl
@@ -114,6 +114,12 @@ function build_runtime(@nospecialize(job::CompilerJob))
     mod
 end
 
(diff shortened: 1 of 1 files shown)


maleadt:
CI failures look related.

vchuravy:
> CI failures look related.

I am not sure? The nightly failure is the same as on `master`? 
Which failure are you looking at?

maleadt:
Yeah sorry, unrelated after all.

PR #794, opened 2026-05-12
Switch to CompilerCaching.jl

maleadt:
Will be a breaking release. Sadly also drops 1.10.


vchuravy:
> Sadly also drops 1.10.

That's unfortunate and will likely mean that we will have to maintain the prior version of GPUCompiler until there is a new LTS.

maleadt:
> will likely mean that we will have to maintain the prior version of GPUCompiler until there is a new LTS.

I'm not convinced that's needed. As long as there's no breaking releases in the back-ends, users can simply use different versions of say CUDA.jl 6.x depending on which Julia version they're using. And for critical features I'd rather they request backports over there rather than maintaining multiple compiler stacks.

vchuravy:
The issue is that for Enzyme we won't be able to drop 1.10 support  (due to DiffEq and so forth) and there the situation is less stable than for the GPU backends.

maleadt:
Grmbl. I'll try to come up with something that keeps 1.10 working then.

PR #799, opened 2026-05-19
Hide GPU runtime stubs from the native cache (v2.0)

maleadt:
Re-lands JuliaGPU/GPUCompiler.jl#749, by forcing stacked method overlays to be used by all back-ends. CPU-visible bodies of the runtime intrinsic stubs no longer reference `extern gpu_*` symbols, so AOT pipelines (`compile=all` sysimages, juliac, PrecompileTools workloads) can link cleanly. The real ccall bodies move into overlays in a private `GLOBAL_METHOD_TABLE`.

Replaces the singular `method_table(job)` override with plural `method_tables(job)` returning a tuple of `Core.MethodTable`s. The default `method_table_view` stacks them on top of `GLOBAL_METHOD_TABLE`, so back-ends inherit the runtime overlays without naming them. `GLOBAL_METHOD_TABLE` is internal and back-ends must not `@overlay` it directly.

Back-end migration: `method_table(::MyJob) = mt` becomes `method_tables(::MyJob) = (mt,)`. Back-ends that previously overrode `method_table_view` to build their own `StackedMethodTable` can usually collapse to a `method_tables` tuple instead.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -298,7 +298,7 @@ pass_by_ref(@nospecialize(job::CompilerJob)) = false
 valid_function_pointer(@nospecialize(job::CompilerJob), ptr::Ptr{Cvoid}) = false
 
 # Care is required for anything that impacts:
-#   - method_table
+#   - method_tables
 #   - inference_params
 #   - optimization_params
 # By default that is just always_inline
@@ -306,11 +306,11 @@ valid_function_pointer(@nospecialize(job::CompilerJob), ptr::Ptr{Cvoid}) = false
 struct GPUCompilerCacheToken
     target_type::Type
     always_inline::Bool
-    method_table::Core.MethodTable
+    method_tables::Tuple{Vararg{Core.MethodTable}}
 end
 
 ci_cache_token(@nospecialize(job::CompilerJob)) =
-    GPUCompilerCacheToken(typeof(job.config.target), job.config.always_inline, method_table(job))
+    GPUCompilerCacheToken(typeof(job.config.target), job.config.always_inline, method_tables(job))
 
 # the codeinstance cache to use -- should only be used for the constructor
 if VERSION >= v"1.11.0-DEV.1552"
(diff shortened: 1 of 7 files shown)


maleadt:
It's kinda unfortunate that this is breaking, but the migration to CompilerCaching.jl is going to be breaking anyway, so let's go with the cleaner API here.

src/interface.jl
```diff
@@ -298,19 +298,19 @@ pass_by_ref(@nospecialize(job::CompilerJob)) = false
...
+#   - method_tables
 #   - inference_params
 #   - optimization_params
 # By default that is just always_inline
 # the cache token is compared with jl_egal
 struct GPUCompilerCacheToken
     target_type::Type
     always_inline::Bool
-    method_table::Core.MethodTable
+    method_tables::Tuple{Vararg{Core.MethodTable}}
```
vchuravy, line 309: Uuf can we use SimpleVector here?

src/interface.jl
```diff
@@ -327,10 +327,38 @@ function ci_cache(@nospecialize(job::CompilerJob))
...
+
+If the back-end has overlays spread across multiple `Core.MethodTable`s (e.g. one local
+to the package plus one inherited from a shared intrinsics library), return them in
+priority order — the first match wins.
+
+For full control of the inference-side `Core.Compiler.MethodTableView`, override
+[`method_table_view`](@ref) instead; that is an internal hook and most back-ends should
+not need it.
+"""
+method_tables(@nospecialize(job::CompilerJob)) = ()
```
vchuravy, line 350: Could we make this non-breaking by defining `method_tables(...) = (method_table(...),)`

src/interface.jl
```diff
@@ -327,10 +327,38 @@ function ci_cache(@nospecialize(job::CompilerJob))
...
+
+# Build the inference-side view of the back-end's method tables stacked on top of
+# GPUCompiler's runtime-intrinsic overlay table. Back-ends generally shouldn't override
+# this; override `method_tables(job)` instead.
+function method_table_view(@nospecialize(job::CompilerJob))
+    parent = CC.OverlayMethodTable(job.world, GLOBAL_METHOD_TABLE)
```
vchuravy, line 356: This isn't an `OverlayMethodTable`,should probably be `InternalMethodTable` or

src/interface.jl
maleadt: We could, but there's going to be a breaking release anyway (probably).

maleadt:
Alternatively, we could avoid this entirely by making the stub runtime functions not break the Julia JIT. But that's probably tricky.

gbaraldi:
We could stub to a donotdelete from llvm? With metadata/call bundles?

maleadt:
Replaced by https://github.com/JuliaGPU/GPUCompiler.jl/pull/808

PR #800, opened 2026-05-19
Add PTXFDivFastPass to lower fdiv fast to NVPTX approximate division

vchuravy:
Overarching goal is to move the fast math handling from CUDA.jl to the GPUCompiler backend.

The LLVM NVPTX backend handles fdiv fast for Float32 (→ div.approx.ftz.f32)
but has no fast path for Float64. This IR-level pass covers both:
- Float32: replaces fdiv with __nv_fast_fdividef (libdevice)
- Float64: replaces fdiv with rcp.approx.ftz.d + Newton refinement,
  matching CUDA.jl's inv_fast(::Float64) algorithm

The pass fires when the instruction carries the afn fast-math flag (set by
@fastmath) or when target.fastmath=true. It follows the NVVMReflectPass
pattern already in ptx.jl.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/ptx.jl
+++ b/src/ptx.jl
@@ -262,10 +262,10 @@ function optimize_module!(@nospecialize(job::CompilerJob{PTXCompilerTarget}),
     # TODO: Use the registered target passes (JuliaGPU/GPUCompiler.jl#450)
     @dispose pb=NewPMPassBuilder() begin
         register!(pb, NVVMReflectPass())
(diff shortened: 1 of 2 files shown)


maleadt:
Rebased and added a Float32 path that plays nicely with #804, so that we can avoid these overrides in CUDA.jl.

PR #801, opened 2026-05-19
PTX: Use LLVM 22 back-end from JLL.

maleadt:
Because having to support Julia 1.10, shipping LLVM 15, is becoming a problem. With NVPTX at least there's a relatively clean split between the middle- and back-end, with most back-end passes being scheduled after the optimization pipeline (only AS inference and range intrinsic lowering is part of the middle-end). So we could simply use `llc` from a more modern LLVM version to generate PTX code.

One problem is probably going to be opaque pointers: LLVM 22 rejects them, while Julia 1.10 and 1.11 still use typed pointers by default.


gbaraldi:
Can't we upgrade to opaque pointers quite easily though? (I think if you parse the typed module in a opaque context and then spit it out should do the trick)

maleadt:
Yeah but that's relatively expensive. In any case, I don't see failures here, which isn't what I expected. Will have to investigate.

maleadt:
Passes all tests on CUDA.jl! At least on 1.12.

PR #803, opened 2026-05-19
Fixes for nightly

maleadt:


maleadt:
Hmm, I can't reproduce these failures locally...

PR #804, opened 2026-05-20
Add a pass to apply fastmath attributes.

maleadt:
GPUCompiler.jl/CUDA.jl are  currently using CUDA's libdevice in a way that's not aligned with LLVM: We use the library whenever possible, setting flags like `__CUDA_PREC_SQRT` to make it behave "precisely" when `fastmath=false` (the default). That's a fine way of doing things, but doesn't align with LLVM, which instead treats `libdevice` as "the fast library", never sets `__CUDA_PREC_SQRT`, and instead uses regular instructions (`@llvm.sqrt`) for the precise versions. This surfaces when switching to LLVM's NVVMReflect pass, https://github.com/JuliaGPU/GPUCompiler.jl/pull/785, so let's align GPUCompiler.jl with LLVM.

To that end, I'll be switching to LLVM intrinsics in CUDA.jl, however, that requires us setting appropriate function/instruction attributes when compiling with `fastmath=true`. Which is what this PR does.

cc @vchuravy

### DIFF
--- a/src/optim.jl
+++ b/src/optim.jl
@@ -1,5 +1,34 @@
 # LLVM IR optimization
 
+"""
+    apply_fastmath!(mod::LLVM.Module)
+
+Apply fast-math semantics to every function definition in `mod` — as if every
+floating-point operation were wrapped in `@fastmath`. Sets `unsafe-fp-math="true"`
+as a function attribute and turns on all fast-math flags on eligible FP
+instructions.
+
+Back-ends should call this from `finish_linked_module!` when their target has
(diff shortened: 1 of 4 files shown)


vchuravy:
Nice this aligns with the direction I am exploring in https://github.com/JuliaGPU/GPUCompiler.jl/pull/800 where the special lowering of `fdiv fast` is moved to the "backend".

maleadt:
Yep; I'll rebase that PR onto this.

PR #808, opened 2026-05-21
Make `gpu_*` runtime stubs CPU-AOT-safe via weak linkage.

maleadt:
Alternative to https://github.com/JuliaGPU/GPUCompiler.jl/pull/799. Instead of using overlay tables to conditionally define runtime methods, add proper CPU-compatible stubs marked `weak` that get overridden by the back-end versions linked in. This should sidestep the whole issue.

### DIFF
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -76,17 +76,81 @@ function compile(def, return_type, types, llvm_return_type=nothing, llvm_types=n
     end
     methods[name] = meth
 
-    # FIXME: if the function is a symbol, implying it will be specified by the target,
-    #        we won't be able to call this function here or we'll get UndefVarErrors.
-    #        work around that by generating an llvmcall stub. can we do better by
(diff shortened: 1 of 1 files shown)


src/runtime.jl
```diff
@@ -76,19 +76,83 @@ function compile(def, return_type, types, llvm_return_type=nothing, llvm_types=n
...
+        ret!(builder, const_inttoptr(ConstantInt(i64, 1), rt))
+    elseif rt isa LLVM.IntegerType
+        ret!(builder, ConstantInt(rt, 0))
+    elseif rt isa LLVM.LLVMFloat || rt isa LLVM.LLVMDouble
+        ret!(builder, ConstantFP(rt, 0.0))
+    else
+        error("Unsupported runtime stub return type: $rt")
+    end
+end
+
```
maleadt, line 155: Pretty questionable, but this code really shouldn't be ever called. Even on the CPU, it's replaced by stronger methods from the CPU runtime library.

PR #810, opened 2026-05-27
Metal: turn device-exception traps into returns to avoid GPU hangs

### LINKED ISSUE #433: Simple throwing kernel hangs
This simple kernel will cause a hang leading to 100% GPU usage and taking down the WindowServer eventually. A similar Metal/C++ kernel runs just fine. The generated IR is also mostly identical.

Another interesting observation is that one can save the WindowServer from its death spiral by running the good C++ kernel.

```julia
using Metal

function kernel()
    throw(nothing)
    return nothing
end

Metal.@sync @metal threads=8 kernel()
```

```c++
void gpu_report_exception() {
    return;
}

void gpu_signal_exception() {
    return;
}

kernel void add_arrays()
{
    gpu_report_exception();
    gpu_signal_exception();

    __builtin_trap();
    __builtin_unreachable();
}
```

maleadt:
On Apple GPUs a compute `trap` wedges the whole device (no compute watchdog; only a reboot clears it), so device-side exceptions hung the GPU on macOS 15+. Run `replace_unreachable!` unconditionally and have it strip the preceding `llvm.trap` and synthesize a return when a function only contains `unreachable`.

Fixes JuliaGPU/Metal.jl#433.


maleadt:
CI failures unrelated.

PR #823, opened 2026-06-03
Make GPUCompiler methods of a PTXCompilerTarget overridable without t…

wsmoses:
…ype piracy

cc @gbaraldi @vchuravy @giordano @maleadt 

since GPUCompiler defines several APIs, like optimization_options, on the job itself -- downstream compilers that use the PTX codegen cannot modify these methods without type piracy on all targets, or vendoring the entire api.

Adding a type parameter to PTXCompilerTarget enables this to work without piracy or vendoring large parts of the codebase.

See https://github.com/EnzymeAD/Enzyme-JAX/issues/2522

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -10,7 +10,7 @@ steps:
       julia -e 'println("--- :julia: Instantiating project")
                 using Pkg
                 Pkg.develop(; path=pwd())
-                Pkg.develop(; name="CUDA")' || exit 3
+                Pkg.develop(; name="CUDA#ptxtarget_ext")' || exit 3
 
       julia -e 'println("+++ :julia: Running tests")
                 using Pkg
--- a/src/ptx.jl
+++ b/src/ptx.jl
@@ -295,10 +295,11 @@ function optimize_module!(@nospecialize(job::CompilerJob{<:PTXCompilerTarget}),
             register!(pb, NVVMReflectPass())
             add!(pb, NVVMReflectPass())
         end
-        add!(pb, PTXRSqrtFastPass())
-        add!(pb, PTXFDivFastPass())
-        add!(pb, PTXFSqrtFastPass())
-
+        if get(optimization_options(job), :ptxfastmath, true)
+            add!(pb, PTXRSqrtFastPass())
+            add!(pb, PTXFDivFastPass())
+            add!(pb, PTXFSqrtFastPass())
+        end
         add!(pb, NewPMFunctionPassManager()) do fpm
             # needed by GemmKernels.jl-like code
             add!(fpm, SpeculativeExecutionPass())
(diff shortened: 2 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -10,7 +10,7 @@ steps:
       julia -e 'println("--- :julia: Instantiating project")
                 using Pkg
                 Pkg.develop(; path=pwd())
-                Pkg.develop(; name="CUDA")' || exit 3
+                Pkg.develop(; name="CUDA#ptxtarget_ext")' || exit 3
```
maleadt, line 13: The whole idea of reverse CI is to see if changes don't break anything. Having it run on an adapted CUDA.jl breaks that set-up.

src/ptx.jl
```diff
@@ -13,7 +13,7 @@ export PTXCompilerTarget
     ArchFeatures     = 2
 end
 
-Base.@kwdef struct PTXCompilerTarget <: AbstractCompilerTarget
+Base.@kwdef struct PTXCompilerTarget{BackendInfo} <: AbstractCompilerTarget
```
maleadt, line 16: The compiler target is specific to GPUCompiler.jl and not intended to be overridden. That's the purpose of the CompilerParams, which is the downstream configurable struct.

src/ptx.jl
```diff
@@ -291,10 +295,11 @@ function optimize_module!(@nospecialize(job::CompilerJob{PTXCompilerTarget}),
...
         end
-        add!(pb, PTXRSqrtFastPass())
-        add!(pb, PTXFDivFastPass())
-        add!(pb, PTXFSqrtFastPass())
-
+        if get(optimization_options(job), :ptxfastmath, true)
+            add!(pb, PTXRSqrtFastPass())
+            add!(pb, PTXFDivFastPass())
+            add!(pb, PTXFSqrtFastPass())
+        end
```
maleadt, line 302: Unrelated.

maleadt (changes requested):
Adding an unused type parameter just for the sake of Reactant is a code smell. Please state the intended outcome here.

src/ptx.jl
wsmoses: will move to separate pr

wsmoses:
@maleadt so we need to modify some of the passes, specifically not use instcombine, and also not run the fast math lowering passes.

In the earlier PR you implemented this as a GPUCompiler.optimization_passes(job) [or similar name]. Reactant uses the existing CUDA job / codegen, so we cannot override this without causing type piracy. Adding a type parameter to the target allows us to override this now and therefore disable instcombine.

Alternatively, if we don't want want to modify the job we can add a field within PTXTarget which contains whether or not to run lowering and/or instcombine, and implement a default   optimization_passes(job{PTXTarget}) which returns a named tuple using those values. That's why I initially had the instcombine flag within the PTXTarget

PR #830, opened 2026-06-04
don't depwarn while holding the typeinf lock

simeonschaub:
This causes a context switch, which I believe is illegal while holding the typeinf lock. This caused hangs in the CoolPDLP.jl test suite on Julia 1.10


maleadt (changes requested):
Please try https://github.com/JuliaGPU/GPUCompiler.jl/pull/831 instead.

PR #831, opened 2026-06-04
Emit deprecation warnings without switching tasks.

maleadt:
`Base.depwarn` logs through the active logger, and the I/O involved can switch tasks. That is illegal while holding the typeinf lock, which our `@locked` compilation entry points do on Julia 1.10/1.11, and caused hangs (#830).

Replace it with `safe_depwarn`, which writes the warning synchronously to `Core.stderr` like `@safe_warn` does. Caller attribution, warning once per call site, and throwing under `--depwarn=error` still behave like `Base.depwarn`; on 1.12+, where compilation doesn't hold the typeinf lock, it just forwards there.

Closes https://github.com/JuliaGPU/GPUCompiler.jl/pull/831

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,20 +1,20 @@
 # Deprecations scheduled for removal in the next major release.
 
 function defs(mod::LLVM.Module)
-    Base.depwarn("`GPUCompiler.defs(mod)` is deprecated; inline `filter(f -> !isdeclaration(f), collect(functions(mod)))`.",
+    safe_depwarn("`GPUCompiler.defs(mod)` is deprecated; inline `filter(f -> !isdeclaration(f), collect(functions(mod)))`.",
                  :defs)
     filter(f -> !isdeclaration(f), collect(functions(mod)))
 end
 
 function decls(mod::LLVM.Module)
-    Base.depwarn("`GPUCompiler.decls(mod)` is deprecated; inline `filter(f -> isdeclaration(f) && !LLVM.isintrinsic(f), collect(functions(mod)))`.",
+    safe_depwarn("`GPUCompiler.decls(mod)` is deprecated; inline `filter(f -> isdeclaration(f) && !LLVM.isintrinsic(f), collect(functions(mod)))`.",
                  :decls)
     filter(f -> isdeclaration(f) && !LLVM.isintrinsic(f), collect(functions(mod)))
 end
 
 link_library!(mod::LLVM.Module, lib::LLVM.Module) = link_library!(mod, [lib])
 function link_library!(mod::LLVM.Module, libs::Vector{LLVM.Module})
-    Base.depwarn("`GPUCompiler.link_library!` is deprecated; call `LLVM.link!(mod, copy(lib))` directly, or `LLVM.link!(mod, lib; only_needed=true)` with a freshly-parsed library.",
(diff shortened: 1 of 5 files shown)


simeonschaub:
Thanks, that's a much better solution! Would you mind also bumping the GPUCompiler version as well as the CUDA compat, so we can fix CoolPDLP CI?

maleadt:
> as well as the CUDA compat

Releasing CUDA.jl is quite a bit of effort, and won't happen right away. Or what do you mean?

simeonschaub:
I mean that the currently released CUDA version does not seem to be compatible with GPUCompiler 1.18. If I install the latest release of CUDA.jl inside an environment and then try to add GPUCompiler#main, I get:

```
    Updating git-repo `https://github.com/JuliaGPU/GPUCompiler.jl.git`
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package CUDACore [bd0ed864]:
 CUDACore [bd0ed864] log:
 ├─possible versions are: 6.0.1 - 6.1.1 or uninstalled
 ├─restricted by compatibility requirements with GPUCompiler [61eb1bfa] to versions: uninstalled
 │ └─GPUCompiler [61eb1bfa] log:
 │   ├─possible versions are: 1.18.0 or uninstalled
 │   └─GPUCompiler [61eb1bfa] is fixed to version 1.18.0
 └─required (without additional version restrictions) by CUDA [052768ef]
   └─CUDA [052768ef] log:
     ├─possible versions are: 0.1.0 - 6.1.0 or uninstalled
     ├─restricted to versions * by project [0dc50556], leaving only versions: 0.1.0 - 6.1.0
     │ └─project [0dc50556] log:
     │   ├─possible versions are: 0.0.0 or uninstalled
     │   └─project [0dc50556] is fixed to version 0.0.0
     ├─restricted by julia compatibility requirements to versions: [2.3.0, 2.5.0 - 6.1.0] or uninstalled, leaving only versions: [2.3.0, 2.5.0 - 6.1.0]
     ├─restricted by compatibility requirements with LLVM [929cbde3] to versions: 5.5.0 - 6.1.0 or uninstalled, leaving only versions: 5.5.0 - 6.1.0
     │ └─LLVM [929cbde3] log:
     │   ├─possible versions are: 0.9.0 - 9.8.2 or uninstalled
     │   └─restricted to versions 9.8.1 - 9 by GPUCompiler [61eb1bfa], leaving only versions: 9.8.1 - 9.8.2
     │     └─GPUCompiler [61eb1bfa] log: see above
     └─restricted by compatibility requirements with GPUCompiler [61eb1bfa] to versions: 6.0.0 - 6.1.0 or uninstalled, leaving only versions: 6.0.0 - 6.1.0
       └─GPUCompiler [61eb1bfa] log: see above
```

Is there a reason CUDA.jl declares such strict compat bounds for GPUCompiler?

maleadt:
Yes, GPUCompiler switched to `llc` from LLVM 22, while older versions were using the in-process one. This prompted retroactively changing compat bounds: https://github.com/JuliaRegistries/General/pull/157146/changes/67cd46766142e16b920dd1aacb0bf2812a712e3b
Normally we don't have to do this, and play by semver, but this was an exception (not worth tagging a breaking release for).

simeonschaub:
Ah, I see! Any ETA when CUDA will be compatible with the latest GPUCompiler again?

simeonschaub:
Since this is a bugfix, should we perhaps backport this to GPUCompiler 1.17?

maleadt:
Yes, backporting would be fine too.

PR #832, opened 2026-06-04
Metal: drop the spurious i1 operand when lowering ctlz/cttz to air.clz/air.ctz

aplavin:
`lower_llvm_intrinsics!`'s `bit_renames` pass renames `llvm.ctlz`/`llvm.cttz` to `air.clz`/`air.ctz` but keeps LLVM's trailing `i1 is_zero_undef`. AIR's `air.clz`/`air.ctz` take only the value operand (and Metal's `leading_zeros`/`trailing_zeros` device overrides already emit the 1-arg form), so a kernel that reaches both sources ends up with a signature clash the bitcode downgrader rejects:

```
error: '@air.clz.i32' defined with type 'i32 (i32)*' but expected 'i32 (i32, i1)*'
```

Fix: drop the operand whose type isn't the result type (the `i1`) for clz/ctz, mirroring the `air.abs` handling just above. No-op for `ctpop`/`bitreverse`.

Reproducer (only Metal needed — fails on 1.16.0…1.18, OK on ≤1.15.x):

```julia
using Metal
f(o) = (@inbounds o[1] = rem(o[1], 3f0) + leading_zeros(reinterpret(Int32, o[1])); return)
Metal.@metal f(Metal.mtl(Float32[1.5f0]))   # => Compilation to AIR failed
```

`rem(::Float32,::Float32)` reaches `Base.rem_internal`, which calls `ctlz_int` → 2-arg `air.clz`; `leading_zeros` → 1-arg `air.clz` via the device override. Regressed in #821.

### DIFF
--- a/src/metal.jl
+++ b/src/metal.jl
@@ -1473,9 +1473,11 @@ function lower_llvm_intrinsics!(@nospecialize(job::CompilerJob), fun::LLVM.Funct
             end
         end
 
-        # integer bit intrinsics: pure renames to AIR's builtin names (same signature, including
-        # the `i1` on clz/ctz). Apple's frontend emits these `air.*` rather than the `llvm.*`
-        # forms, so we rename rather than rely on the metallib loader accepting `llvm.*`.
+        # integer bit intrinsics: rename to AIR's builtin names. ctpop/bitreverse keep their
+        # single value operand; clz/ctz must drop LLVM's trailing `i1 is_zero_undef`, since
+        # AIR's air.clz/air.ctz take only the value operand. Apple's frontend emits these
+        # `air.*` rather than the `llvm.*` forms, so we rename rather than rely on the metallib
+        # loader accepting `llvm.*`.
         bit_renames = Dict(
             LLVM.Intrinsic("llvm.ctlz")       => ("llvm.ctlz",       "air.clz"),
             LLVM.Intrinsic("llvm.cttz")       => ("llvm.cttz",       "air.ctz"),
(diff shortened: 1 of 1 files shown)


aplavin:
disclaimer: llm-generated under my detailed supervision, that's why marking as draft
the bug is real, the fix seems to work, but I'm not familiar with the codebase to understand if it's a clean fix or not

aplavin:
Reproducer:
```julia
using Metal
f(o) = (@inbounds o[1] = rem(o[1], 3f0) + leading_zeros(reinterpret(Int32, o[1])); return)
Metal.@metal f(Metal.mtl(Float32[1.5f0]))
```

maleadt:
> AIR's `air.clz`/`air.ctz` take only the value operand (and Metal's `leading_zeros`/`trailing_zeros` device overrides already emit the 1-arg form)

That's not true; Apple also emits the `i1`:

```llvm
declare i32 @air.clz.i32(i32, i1)          ; scalar
declare <4 x i32> @air.clz.v4i32(<4 x i32>, i1)  ; vector
declare i32 @air.abs.s.i32(i32)            ; abs really is 1-arg
```

The underlying problem is that we get two conflicting definitions of this intrinsic in the module, an old one from Metal.jl, and a new one with trailing `i1` from GPUCompiler.jl. I think the proper solution is to prevent newer GPUCompiler.jl's from getting installed together with older Metal.jl's, as from https://github.com/JuliaGPU/Metal.jl/pull/801 on everything is aligned with https://github.com/JuliaGPU/GPUCompiler.jl/pull/821.

EDIT: https://github.com/JuliaRegistries/General/pull/157199

aplavin:
Thanks! So basically I caught a transient error because of Metal vs GPUCompiler release timings mismatch, that's it?

maleadt:
Yeah, sorry for that.

PR #836, opened 2026-06-06
PTX: modernize compilation now that LLVM 22 is used

maleadt:
PTX machine code is now generated by an external, up-to-date LLVM 22; the in-process LLVM only drives the middle end. This PR ports several improvements that build on that split:

- the middle end now uses a generic `TargetMachine`, folds `__nvvm_reflect` itself at the true target, and accepts NVVM intrinsics the in-process LLVM does not know about. Together this fully supports devices newer than the in-process LLVM;
- the datalayout, launch-bound attributes, and kernel calling convention now match what the external back-end expects;
- workarounds the back-end no longer needs are gone: `lower_unreachable!`, byval lowering of arguments it wouldn't copy, and fast-math rewrites of f32 operations it handles natively;
- warnings and errors from the external `llc` are now surfaced.


maleadt:
CI failure is benign; we're actually generating better code now.

PR #838, opened 2026-06-08
Revert "PTX: only lower byval arguments the back-end would copy."

maleadt:
This reverts commit 189c66eea5dd4a5b98bc4b8a8b1fe3cb2b69129d. Turns out byval lowering in NVPTX isn't good enough yet, and hurts performance.


maleadt:
Debugged and fixed upstream: https://github.com/llvm/llvm-project/pull/202379

PR #842, opened 2026-06-09
Gate aggressive instcombine pass

wsmoses:
we're going to need to backport this to be compatible with cuda5 too, so 1.17

cc @giordano

### DIFF
--- a/src/optim.jl
+++ b/src/optim.jl
@@ -40,6 +40,15 @@ function instcombine_pass(@nospecialize(job::CompilerJob))
     end
 end
 
+# Pick the peephole pass according to `optimization_options(job).aggressiveinstcombine`.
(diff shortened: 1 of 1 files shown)


src/optim.jl
```diff
@@ -40,6 +40,15 @@ function instcombine_pass(@nospecialize(job::CompilerJob))
     end
 end
 
+# Pick the peephole pass according to `optimization_options(job).aggressiveinstcombine`.
+function aggressiveinstcombine_pass(@nospecialize(job::CompilerJob))
+    if get(optimization_options(job), :aggressiveinstcombine, true)
```
maleadt, line 45: Does this really need another flag? Isn't this in the same vein as the existing `:instcombine` one?

wsmoses: they are different passes and do different things. So since it's not breaking to add another key, I'd prefer a different pass name

PR #844, opened 2026-06-10
Record line coverage of device code at compile time.

maleadt:
When the host process runs with --code-coverage, lowered code contains :code_coverage_effect expressions, which the compiler only inserts for code that is actually being tracked. Device code cannot count executions (it cannot call into the Julia runtime), so instead mark those lines as visited when compiling them, by calling jl_coverage_visit_line on the host. Coverage of device code thus means "this code was compiled", with counts reflecting the number of compilations rather than executions.

Inspired by https://github.com/JuliaGPU/cuTile.jl/pull/249

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -707,6 +707,37 @@ function Base.precompile(@nospecialize(job::CompilerJob))
 end
 
 
+## code coverage
+
+# When the host process runs with `--code-coverage`, lowered code contains
+# `:code_coverage_effect` expressions, which the compiler only inserts for code that is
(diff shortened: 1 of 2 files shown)


vchuravy:
Well, we could actually do this during runtime ;) We just need an atomic increment to a linetable buffer and then after the kernel is done we report those to Julia :) 

But I think this variant is a good first step and will make @sloede very happy.

maleadt:
> Well, we could actually do this during runtime ;) We just need an atomic increment to a linetable buffer and then after the kernel is done we report those to Julia :)

I guess, but it's much more involved... So yeah, good first step :-)

PR #846, opened 2026-06-12
Update Buildkite pipeline for the new JuliaGPU cluster

maleadt:
The JuliaGPU Buildkite agents have moved to a dedicated cluster, where the single `juliagpu` queue has been split into per-backend queues. Steps now select agents using `queue: "cuda"`, `queue: "rocm"` or `queue: "oneapi"` instead of `queue: "juliagpu"` combined with a `cuda`/`rocm`/`intel` tag.

Since queues are scoped to a cluster, a single pipeline can no longer mix JuliaGPU and `juliaecosystem` jobs: `pipeline.yml` now contains only the GPU jobs (and remains the default upload for the existing Buildkite pipeline), while the `juliaecosystem` steps (Metal.jl, Enzyme.jl) have moved unmodified to `pipeline-julia.yml`, to be uploaded by a separate pipeline in the ecosystem cluster once that is back online.

### DIFF
--- a/.buildkite/pipeline-julia.yml
+++ b/.buildkite/pipeline-julia.yml
@@ -1,0 +1,51 @@
+steps:
+  - label: "Metal.jl"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1.11"
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    command: |
+      julia -e 'println("--- :julia: Instantiating project")
+                using Pkg
+                Pkg.develop(; path=pwd())
+                Pkg.develop(; name="Metal")' || exit 3
+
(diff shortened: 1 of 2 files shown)


vchuravy (approved):

vchuravy:
That means we need to update the webui to upload two pipeline*yml?

Ah if I read the docs correctly we now need two buildkite pipelines? One in the GPU cluster, the other in the ecosystem cluster?

maleadt:
> Ah if I read the docs correctly we now need two buildkite pipelines? One in the GPU cluster, the other in the ecosystem cluster?

Correct. Sadly. But I don't see a way around, and it was honestly kinda questionable to have these two entirely unrelated CI pipelines get mixed (at least from the agent's perspective).

PR #850, opened 2026-06-17
Fix handling of Julia's non IPO-safe metadata

maleadt:
Julia can emit invariant-load and const-region alias metadata for by-pointer aggregate loads. When GPUCompiler materializes byval arguments or Metal's late inliner exposes those pointers as stack-derived, that metadata becomes false and lets LLVM ignore stack stores; strip the full const-region metadata package in those cases and cover the post-inline alloca shape with a regression test.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GPUCompiler"
 uuid = "61eb1bfa-7361-4325-ad38-22787b887f55"
-version = "1.22.0"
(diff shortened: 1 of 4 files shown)


maleadt:
Actually, reading https://github.com/JuliaLang/julia/issues/44285#issuecomment-2094254055 we should be able to do without this, as long as we only inline late.

PR #857, opened 2026-06-22
GCN: generate machine code via external AMDGPU_LLVM_Backend_jll

vchuravy:
Mirror the `NVPTX_LLVM_Backend_jll` approach (see `src/ptx.jl`) for AMDGPU: generate GCN machine code through the external, up-to-date `llc` from [`AMDGPU_LLVM_Backend_jll`](https://github.com/JuliaRegistries/General/pull/158674) rather than the in-process LLVM back-end.

## Changes
- **`src/gcn.jl`**: add an `AMDGPU_LLVM_Backend_jll` `LazyModule` and an `@unlocked mcgen(::CompilerJob{GCNCompilerTarget}, …)` override that writes the module to a `.bc`, runs the external `llc` (`-mtriple`/`-mcpu=dev_isa`/`-mattr=features`/`--relocation-model=pic`/`-filetype`), surfaces stderr diagnostics, and reads back the asm/object.
- **`Project.toml`**: `AMDGPU_LLVM_Backend_jll` weakdep + `compat = "22"`.
- **`test/Project.toml`**: add as a test dep.
- **`test/runtests.jl`**: load it and skip the `gcn` testsuite when `!AMDGPU_LLVM_Backend_jll.is_available()`.

## Notes
- Unlike PTX, this keeps GCN's `llvm_machine`/`llvm_datalayout` as-is, so the **middle end still requires the in-process AMDGPU back-end** (`:AMDGPU in LLVM.backends()`, already guarded by `test/gcn.jl`); only final machine-code emission goes through the external `llc`. Full decoupling would require a hardcoded datalayout, which is awkward because AMDGPU's datalayout already carries non-integral spaces (`…-ni:7:8:9`) that GPUCompiler then extends with `-ni:10:11:12:13`.
- Depends on the registration of `AMDGPU_LLVM_Backend_jll` (JuliaRegistries/General#158674); until that merges the project will not resolve.

### DIFF
--- a/src/gcn.jl
+++ b/src/gcn.jl
@@ -4,7 +4,6 @@ const AMDGPU_LLVM_Backend_jll =
     LazyModule("AMDGPU_LLVM_Backend_jll",
                UUID("cc5c0156-bd05-5a77-8a68-bb0aafb29019"))
 
-
 ## target
 
 export GCNCompilerTarget
@@ -165,12 +164,6 @@ end
     target = job.config.target
 
     if target.backend === :inprocess
-        # the in-process LLVM back-end is deprecated and will be removed in the next
-        # breaking release in favor of AMDGPU_LLVM_Backend_jll (backend=:external).
-        safe_depwarn(
-            "Generating GCN machine code with the in-process LLVM is deprecated; " *
-            "load AMDGPU_LLVM_Backend_jll and use `backend=:external` instead.",
-            :mcgen)
         if :AMDGPU ∉ LLVM.backends()
             error("The in-process LLVM lacks the AMDGPU target; cannot compile to GCN. " *
                   "Load AMDGPU_LLVM_Backend_jll and use `backend=:external` instead.")
--- a/src/gcn.jl
+++ b/src/gcn.jl
@@ -161,7 +161,18 @@ end
 @unlocked function mcgen(@nospecialize(job::CompilerJob{GCNCompilerTarget}),
                          mod::LLVM.Module, format=LLVM.API.LLVMAssemblyFile)
     if !isavailable(AMDGPU_LLVM_Backend_jll) || !AMDGPU_LLVM_Backend_jll.is_available()
-        error("AMDGPU LLVM back-end not loaded; cannot compile to GCN.")
+        # fall back to the in-process LLVM back-end, which is deprecated and will be
+        # removed in the next breaking release in favor of AMDGPU_LLVM_Backend_jll.
+        safe_depwarn(
+            "Generating GCN machine code with the in-process LLVM is deprecated; " *
+            "load AMDGPU_LLVM_Backend_jll to use the external back-end instead.",
+            :mcgen)
+        if :AMDGPU ∉ LLVM.backends()
+            error("AMDGPU LLVM back-end not loaded and the in-process LLVM lacks the " *
+                  "AMDGPU target; cannot compile to GCN.")
+        end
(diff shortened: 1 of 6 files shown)


vchuravy:
@maleadt did you do anything special to stop older version of CUDA to use GPUCompiler without NVPTX_LLVM_Backend_jll being loaded?

maleadt:
> did you do anything special to stop older version of CUDA to use GPUCompiler without NVPTX_LLVM_Backend_jll being loaded?

Add a compat bound to the registry. Looking back at the problems that caused, I think it would be better to temporarily support both the in-process and external LLVM path, deprecating the former until we cut a breaking release.

test/gcn.jl
```diff
@@ -442,5 +442,32 @@ end
...
+    end
+
+    # simulate AMDGPU_LLVM_Backend_jll not being loaded
+    pkg = Base.PkgId(Base.UUID("cc5c0156-bd05-5a77-8a68-bb0aafb29019"),
+                     "AMDGPU_LLVM_Backend_jll")
+    saved = get(Base.loaded_modules, pkg, nothing)
+    try
+        delete!(Base.loaded_modules, pkg)
+        @test !GPUCompiler.isavailable(GPUCompiler.AMDGPU_LLVM_Backend_jll)
+
```
vchuravy: @maleadt this is a fairly horrifying way of testing this. Any better ideas?

src/gcn.jl
```diff
@@ -1,14 +1,21 @@
 # implementation of the GPUCompiler interfaces for generating GCN code
 
+const AMDGPU_LLVM_Backend_jll =
+    LazyModule("AMDGPU_LLVM_Backend_jll",
+               UUID("cc5c0156-bd05-5a77-8a68-bb0aafb29019"))
+
+
```
vchuravy:
```suggestion
```

src/gcn.jl
```diff
@@ -153,6 +160,76 @@ function add_kernarg_address_spaces!(
...
+                         mod::LLVM.Module, format=LLVM.API.LLVMAssemblyFile)
+    target = job.config.target
+
+    if target.backend === :inprocess
+        # the in-process LLVM back-end is deprecated and will be removed in the next
+        # breaking release in favor of AMDGPU_LLVM_Backend_jll (backend=:external).
+        safe_depwarn(
+            "Generating GCN machine code with the in-process LLVM is deprecated; " *
+            "load AMDGPU_LLVM_Backend_jll and use `backend=:external` instead.",
+            :mcgen)
```
vchuravy:
```suggestion
```

test/gcn.jl
```diff
@@ -3,6 +3,37 @@ if :AMDGPU in LLVM.backends()
...
+    # the backend participates in the runtime slug, so different back-ends don't share a cache
+    job_ext, _ = GCN.create_job(mod.kernel, Tuple{}; backend=:external)
+    job_inp, _ = GCN.create_job(mod.kernel, Tuple{}; backend=:inprocess)
+    @test endswith(GPUCompiler.runtime_slug(job_ext), "-external")
+    @test endswith(GPUCompiler.runtime_slug(job_inp), "-inprocess")
+    @test GPUCompiler.runtime_slug(job_ext) != GPUCompiler.runtime_slug(job_inp)
+
+    # the explicit :external backend generates machine code through the external llc
+    @test (GCN.code_native(devnull, mod.kernel, Tuple{}; backend=:external); true)
```
vchuravy, line 31: Add the same test for `:internal`

PR #859, opened 2026-06-22
Add `alloca` intrinsic for per-workitem stack scratch

vchuravy:
## Summary

Adds `GPUCompiler.alloca(::Type{T}, ::Val{N})::Ptr{T}`, a primitive that hands device code a fixed-size, per-workitem stack scratch buffer for `N` elements of `T`. The intended use is to replace abstractions like KernelAbstractions' `@private` `MArray`-backed scratchpad with a direct stack allocation (see companion KA PR).

## Why an intrinsic instead of `llvmcall` + `alloca`

Emitting the `alloca` directly through `llvmcall` is unsound/ineffective:

1. **Promotion is lost.** A `Ptr` round-trips through `ptrtoint`/`inttoptr` at the `llvmcall` boundary, so SROA/mem2reg can never promote the slot to registers — which is the whole point of small `@private` scratch.
2. **Address space isn't known at the front end.** On NVPTX/AMDGPU an `alloca` must live in the datalayout's alloca address space (e.g. addrspace 5) and be `addrspacecast` to generic; this can't be written portably across the PTX/GCN/Metal/SPIR-V back-ends from Julia.
3. **Lifetime.** Per the LangRef, an `alloca`'s storage is freed when *its* function returns; in an `llvmcall` wrapper that's the wrapper, and correctness relies entirely on the inliner relocating a static entry-block alloca.

## How it works

- The front end emits a declared-only `julia.gpu.alloca.<bytes>.<align>` intrinsic (modeled on `julia.gpu.debug_level`).
- `lower_alloca!`, run from `irgen` *before* the optimizer, materializes it as a real entry-block `alloca` in the datalayout's alloca address space (so SROA/mem2reg can promote it), cast back to generic for the returned `Ptr`. This produces the same IR shape Julia already emits for mutable stack allocations.
- `T` must be `isbits`; a zero-byte request returns a null pointer.

## Testing

Adds a native-target testset asserting an entry-block `alloca [N x i8], align <align>`, full promotion when optimized, no surviving intrinsic, and the zero-byte → null path.

Verified end-to-end (including execution) through the POCL/SPIR-V back-end via the companion KernelAbstractions change: `@private Float32 (4,)` lowers to `alloca [16 x i8], align 4` in addrspace 0 (OpenCL `Function`/private) and runs correctly.

### DIFF
--- a/src/irgen.jl
+++ b/src/irgen.jl
@@ -1242,7 +1242,8 @@ function alloca_intr(mod::LLVM.Module, bytes::Integer, align::Integer)
     else
         # returns an opaque pointer; intentionally *not* readnone/speculatable, as each call
         # must yield a distinct slot and must not be hoisted or CSE'd.
-        LLVM.Function(mod, name, LLVM.FunctionType(LLVM.PointerType()))
+        T_ptr = LLVM.supports_typed_pointers(LLVM.context(mod)) ? LLVM.PointerType(LLVM.Int8Type()) : LLVM.PointerType()
+        LLVM.Function(mod, name, LLVM.FunctionType(T_ptr))
     end
     return intr
 end
@@ -1264,7 +1265,7 @@ function alloca_value(@nospecialize(T), N::Int)
     end
 
     @dispose ctx=Context() begin
-        T_ptr = LLVM.PointerType()
+        T_ptr = LLVM.supports_typed_pointers(ctx) ? LLVM.PointerType(LLVM.Int8Type()) : LLVM.PointerType()
 
         # create function
         llvm_f, _ = create_function(T_ptr)
@@ -1321,10 +1322,14 @@ function lower_alloca!(@nospecialize(job::CompilerJob), mod::LLVM.Module)
 
                 # `alloca!` placed the slot in the datalayout's alloca address space; cast back
                 # to generic (AS 0) to match the `Ptr` `alloca` returns.
-                ptr = if LLVM.addrspace(value_type(slot)) == 0
+                ptr = if value_type(slot) == value_type(call)
                     slot
                 else
-                    addrspacecast!(builder, slot, LLVM.PointerType())
+                    if LLVM.addrspace(value_type(slot)) == LLVM.addrspace(value_type(call))
+                        bitcast!(builder, slot, value_type(call))
+                    else
+                        addrspacecast!(builder, slot, value_type(call))
+                    end
                 end
                 replace_uses!(call, ptr)
             end
--- a/src/irgen.jl
+++ b/src/irgen.jl
@@ -149,6 +149,10 @@ function irgen(@nospecialize(job::CompilerJob))
         # the job's configured level, so device code can branch on it as a compile-time
         # constant that is part of the cache key (unlike reading the `-g` global directly).
         lower_debug_level!(job, mod)
+
+        # materialize `GPUCompiler.alloca` intrinsics as real entry-block allocas, before the
+        # optimizer runs so the slots can be promoted (see `lower_alloca!`).
+        lower_alloca!(job, mod)
     end
 
     return mod, compiled, gv_to_value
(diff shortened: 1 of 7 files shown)


maleadt:
Conceptually fine, but I didn't realize we needed this.

> * **Promotion is lost.** A `Ptr` round-trips through `ptrtoint`/`inttoptr` at the `llvmcall` boundary, so SROA/mem2reg can never promote the slot to registers — which is the whole point of small `@private` scratch.

Hmm, isn't this optimized away?
 
> * **Address space isn't known at the front end.** On NVPTX/AMDGPU an `alloca` must live in the datalayout's alloca address space (e.g. addrspace 5) and be `addrspacecast` to generic; this can't be written portably across the PTX/GCN/Metal/SPIR-V back-ends from Julia.

Yeah, I guess. How do we make sure Julia's `alloca`'s are in the correct AS? Couldn't/Shouldn't we patch those up afterwards?

> * **Lifetime.** Per the LangRef, an `alloca`'s storage is freed when _its_ function returns; in an `llvmcall` wrapper that's the wrapper, and correctness relies entirely on the inliner relocating a static entry-block alloca.

`llvmcall` wrappers are `alwaysinline`; doesn't that suffice?

vchuravy:
The crux is that llvmcall + alloca is UB and I have seen LLVM delete the alloca and replace it with poison. I can dig up the old PR I had for that.

My overarching goal is to remove the StaticArrays dependency that Kernel abstraction currently forces and which you have complained about before since creates behavior where we have to hope for escape analysis to work. 

The rest is Claude obsessing over details. We have fixed alloca Julia to be emitted correctly

src/irgen.jl
```diff
@@ -1215,6 +1219,125 @@ function lower_debug_level!(@nospecialize(job::CompilerJob), mod::LLVM.Module)
...
+    N >= 0 || throw(ArgumentError("GPUCompiler.alloca count must be non-negative, got $N"))
+
+    bytes = sizeof(T) * N
+    align = Base.datatype_alignment(T)
+
+    # a zero-byte allocation has no storage to point at; hand back a null pointer rather than
+    # emitting a degenerate 0-element alloca.
+    if bytes == 0
+        return :(reinterpret(Ptr{$T}, C_NULL))
+    end
```
vchuravy, line 1269: Should we return a `Base.LLVMPtr`

vchuravy: We can't without adding it to the generated function as an input argument.

gbaraldi:
https://github.com/JuliaLang/julia/blob/3d710776cb31ce05775ee10290f709df1f1e939c/src/codegen.cpp#L2308 This is how base does it for the alloca address-ace

test/Project.toml
```diff
@@ -23,3 +23,6 @@ demumble_jll = "1e29f10c-031c-5a83-9565-69cddfc27673"
 Aqua = "0.8"
 LLVM_jll = "15,16,18,20"
 ParallelTestRunner = "2"
+
+[sources]
+GPUCompiler = { path=".." }
```
maleadt: This shouldn be needed, right? Within the workspace, GPUCompiler should resolve correctly. Or is this for 1.11?

vchuravy:
```suggestion
```

gbaraldi:
https://github.com/JuliaLang/julia/blob/3d710776cb31ce05775ee10290f709df1f1e939c/src/codegen.cpp#L2316-L2328 for the size of the alloys (we may also just want to do always power of 2)

gbaraldi:
The one thing I'm a little worried is using Ptr{} as a representation. Those addrspacecasts aren't free on amdgpu. addrspace(0) has an addressing cost when translating to private memory so it might be worth to check if the codegen of it is as expected

vchuravy:
I sadly can't get around it. We don't know the target during the generated function stage, but we must provide a stable return type so I can't use LLVMPtr

gbaraldi:
Can't we make a type for this? Something like AllocaArray or something? Though I guess it needs more intrinsics then.

Another option is to just remove the addrspacecast in a pass? 

Edit: In theory InferAddressSpaces is capable of removing these, but it deserves a double check (I'm not sure if our 64 bit indexing mucks that up)

vchuravy:
> Can't we make a type for this? 

Open to all ideas, but I am essentially working around the fact here that Base Julia has no true support for alloca, and using StaticArrays is just "make a type" 

We can put the AS into the alloca intrinsic as a type parameter and then force each backend to use a device overlay

vchuravy:
@gbaraldi I think I am happy with this. Since each backend will have to define `Scratchspace` already we can add the AS target knowledge into the intrinsic.

PR #864, opened 2026-06-26
Turn Julia passes into closures

vchuravy:
I was thinking about `current_job` and if we could replace it,
and I think if we turn our passes into manual closures then we could pass it in, instead of relying on a global variable.


maleadt:
Oh, that's cute! And would have avoided bugs like the one fixed in https://github.com/JuliaGPU/GPUCompiler.jl/pull/855.
I've made Claude apply it to the other passes.

maleadt:
Nightly failure unrelated.

PR #865, opened 2026-06-27
compiler: Use upcoming TypeEgal dispatch keys for type callees

Keno:
The Base PR JuliaLang/julia#62001 is expected to spell exact closed type-object dispatch keys as Core.TypeEgal. When looking up method instances from a stable function type, use that upcoming dispatch spelling for closed Type{T} callee keys so constructor calls such as Memory{Int}(undef, n) remain compatible with Base.isdispatchtuple.

AI disclosure: This commit was prepared with assistance from generative AI. The resulting PR should be checked carefully by a maintainer before merging.

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -16,9 +16,21 @@ end
 
 export methodinstance, generic_methodinstance
 
+@static if isdefined(Core, :TypeEgal)
+    @inline function dispatch_function_key(ft::Type)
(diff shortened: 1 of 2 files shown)


maleadt:
Subsumed by https://github.com/JuliaGPU/GPUCompiler.jl/pull/871 (I couldn't push here)

PR #874, opened 2026-07-10
Materialize isbits boxed constants as device globals

maleadt:
Since JuliaLang/julia#55045 (1.14.0-DEV.1348), small isbits unions built from
constants are kept fully boxed: codegen emits a `julia.constgv` slot holding a
pointer to the *host* box, which device code then dereferences for the payload
or type tag. `relocate_gvs!` used to bake that host address into the IR, so any
such kernel reads host memory → `ERROR_ILLEGAL_ADDRESS`.

Fixing this upstream was rejected (JuliaLang/julia#62191): inline storage is a
host-side de-optimization, and flag-guarding every eager-box site doesn't
scale. Instead, this implements @topolarity's suggestion on the GPUCompiler
side, where the GC-related objections don't apply.

`relocate_gvs!` now inspects the object behind each constgv slot. Non-ghost,
non-`Bool` isbits objects are *materialized*: a private constant replica of
the box (header word + payload bytes, heap-compatible layout and alignment) is
emitted into the module and the slot points at its payload. Everything else —
types, symbols, singletons, `Bool`s, mutable objects — keeps the baked address;
those are only ever used as identity tokens.

Notes:
- The replica's header holds the same value codegen compares against in
  `typeof`/`isa` (smalltag immediate or type pointer), and `===` on isbits
  falls back to bits comparison, so no identity-sensitive behavior changes.
- Resolution runs before `optimize!`, so payload and tag loads constant-fold;
  the device code ends up identical to what inline union storage produced
  before #55045. No de-optimization, host codegen untouched.
- Every eager-boxing site funnels through `literal_pointer_val_slot`, so
  future upstream boxing patterns are covered automatically.
- Session-dependence marking is now precise: `relocate_gvs!` reports whether
  any host address survived (baked slot, or a materialized header carrying a
  non-smalltag type pointer). Kernels whose constants are all smalltag
  (`Int`/`UInt`/`Char` etc.) no longer get wiped from package images.

Applies to 1.13+ (older versions bake addresses in Julia itself); the union
regression only exists on 1.14.0-DEV.1348+.

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -759,32 +759,100 @@ end
 """
     relocate_gvs!(mod::LLVM.Module, gv_to_value::Dict{String, Ptr{Cvoid}})
 
-Bake absolute pointer values into the initializers of `julia.constgv`-tagged
-global variables, matching them by name against `gv_to_value`. Only GVs that
-are declarations (as produced by `compile_method_instance`, which strips the
-initializers for session portability) or still have a null initializer are
-touched: preserving existing initializers covers the older-Julia path where
-Julia itself emits pointer values directly.
+Resolve `julia.constgv`-tagged global variable slots, matching them by name
+against `gv_to_value`. Slots whose object is a non-ghost, non-`Bool` isbits
+value are *materialized*: a device-resident constant replica of the box
+(header word + payload bytes) is emitted and the slot points at its payload,
+so device code can dereference it and LLVM can constant-fold it. All other
+slots get the object's absolute host address baked in, as before; device code
+only uses those as opaque identity tokens.
+
+Only GVs that are declarations (as produced by `compile_method_instance`,
+which strips the initializers for session portability) or still have a null
+initializer are touched: preserving existing initializers covers the
+older-Julia path where Julia itself emits pointer values directly.
 
 GVs present in `mod` but missing from `gv_to_value` remain declarations, which
 back-ends will reject loudly (undefined symbol) rather than silently folding
 to null.
+
+Returns `true` if the module is session-portable afterwards: no absolute host
+address was written (neither a baked slot nor a materialized header carrying
+a non-smalltag type pointer).
 """
 function relocate_gvs!(mod::LLVM.Module, gv_to_value::Dict{String, Ptr{Cvoid}})
+    portable = true
     mod_gvs = globals(mod)
     for (name, init) in gv_to_value
         haskey(mod_gvs, name) || continue
         gv = mod_gvs[name]
         cur = initializer(gv)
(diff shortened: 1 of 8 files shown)


src/jlgen.jl
```diff
@@ -755,32 +755,100 @@ end
...
+            portable = false
+            continue
+        end
+
+        val = nothing
+        if init != C_NULL
+            obj = Base.unsafe_pointer_to_objref(init)
+            # ghosts/singletons and Bools are compared by pointer; keep their
+            # identity by baking the host address (they are never dereferenced)
+            if isbitstype(typeof(obj)) && sizeof(obj) > 0 && !(obj isa Bool)
```
topolarity, line 800: I assume that for `Bool` you have another solution?

It's not clear to me whether something like this is problematic:
```julia
mod = @eval module $(gensym())
    @noinline produce(cond::Bool, a::Int32) = cond ? a : true
    function consume(cond::Bool, a::Int32)
        x = produce(cond, a)
        x isa Bool && x && return Int32(1)
        return Int32(0)
    end
end
```

where `produce` eagerly returns a boxed `Bool`

topolarity (commented):
Nice! This ended up even simpler than I hoped 👍 

This seems like the better path to me. Something like https://github.com/JuliaLang/julia/pull/62191 is possible, but it requires a more significant refactor to give any assurances that it's complete. This is complete by design, assuming it works, so that's quite nice.

Only concern is `Bool`, which might be worthy of more test coverage.  Some notes inline.

src/jlgen.jl
topolarity: I asked Claude to explore one solution: [singleton_unification_demo.jl.txt](https://github.com/user-attachments/files/29984516/singleton_unification_demo.jl.txt)

Almost unreviewed by me, but I think it got the idea. Hopefully I am not sending garbage your way.

src/jlgen.jl
```diff
@@ -755,32 +755,125 @@ end
...
+
+        # Stay conservative if Bool stops using a smalltag.
+        portable &= hdr < UInt(64 << 4)   # jl_max_tags << 4
+    end
+    return portable
+end
+
+# emit a device-resident constant replica of the box holding `obj`; returns
+# the constant to store in the slot, and the (gcbits-masked) header word
+function materialize_box!(mod::LLVM.Module, gv::GlobalVariable, @nospecialize(obj),
```
vchuravy, line 846: Should we have some safety that `obj` is immutable? Otherwise it isnt trivially copyable?

maleadt: Should be guaranteed by the callers already, but I'll add an assert

PR #878, opened 2026-07-14
Make generated code relocatable across Julia sessions

maleadt:
Keep Julia values and libjulia globals symbolic until final backend lowering, instead of embedding session-specific host addresses in cached IR.

Relocation records are deterministic and remain attached while modules are linked and optimized. A backend selects one of three delivery strategies:

- `:bake` resolves addresses during final emission. This preserves compatibility but keeps results session-local.
- `:patch` emits named writable globals for a loader such as CUDA or ORC to patch after loading.
- `:table` rewrites references through a runtime table for loaders such as Metal that cannot look up data symbols.

Cached backend results persist only when Julia exposes relocatable global metadata and the backend selects `:patch` or `:table`. Resolved Julia values are permanently rooted, runtime bitcode remains symbolic until it reaches the final module, and manifests are frozen after lowering so metadata cannot diverge from emitted code.

The implementation is exercised by CUDA.jl#3200, Metal.jl#916, and JuliaLang/AllocCheck.jl#113. Backends that have not opted in retain the eager `:bake` behavior; this was validated with OpenCL.jl on both its SPIR-V and OpenCL C paths.

Supersedes #125 and #348.


src/interface.jl
```diff
@@ -506,6 +506,49 @@ end
 
 ## session-dependent results
 #
+struct JuliaValueRef
+    value::Any
+end
+
+struct RuntimeGlobalRef
+    name::Symbol
+end
```
vchuravy: I was a bit confused by the name here since I would have expected a GLobalRef to have a module + name, whereas this seems to be a cglobal?

maleadt: Yeah the name is TBD; this is a ref to the runtime library.

maleadt:
Okay, this is starting to look good. IR should be fully relocatable after this, at least for back-ends supporting run-time relocations.

src/optim.jl
```diff
@@ -325,7 +326,9 @@ function buildIntrinsicLoweringPipeline(mpm, @nospecialize(job::CompilerJob), op
             add!(fpm, GPULowerGCFramePass(job))
         end
         if job.config.libraries
-            add!(mpm, GPULinkRuntimePass(job))
+            # The pass builder resolves this name to the registered instance, which
+            # captures the Relocations object owned by optimize!.
+            add!(mpm, "GPULinkRuntime")
             add!(mpm, GPULinkLibrariesPass(job))
             add!(mpm, GPUFinishRuntimeIntrinsicsPass(job))
```
vchuravy: Should we also do this for the `job` capturing?

src/relocation.jl
```diff
@@ -0,0 +1,522 @@
...
+
+"""
+    CGlobalRef(symbol)
+
+A named libjulia C data global. Resolution returns the word stored in that global in the
+current Julia process.
+"""
+struct CGlobalRef
+    symbol::Symbol
+end
```
vchuravy, line 41: Could we generalize that to `libname` + `symbol`

src/relocation.jl
```diff
@@ -0,0 +1,522 @@
...
+    resolve_relocation_target(target) -> UInt
+
+Resolve a relocation target to its word in the current Julia process.
+"""
+function resolve_relocation_target(target::JuliaValueRef)
+    box = Any[target.value]
+    GC.@preserve box begin
+        return unsafe_load(Base.unsafe_convert(Ptr{UInt}, pointer(box)))
+    end
+end
```
vchuravy, line 82: I think the previous version of this uses `jl_value_ptr` directly?

Why the 1-element array instead of a `Ref`?

We had something similar in Enzyme, but we restricted it to only `types`. https://github.com/EnzymeAD/Enzyme.jl/blob/4c8b1c0a04689cce931bd4ad8aa02f6a0919cecc/src/utils.jl#L1-L52

I think for all other cases one can use `pointer_from_objref`.

src/relocation.jl
```diff
@@ -0,0 +1,522 @@
...
+function resolve_relocation_target(target::CGlobalRef)
+    address = ccall(:jl_cglobal, Any, (Any, Any), String(target.symbol), UInt)
```
vchuravy: You dont need to convert to a String here, and can pass the Symbol directly.

vchuravy: I think there is another weirdness here, looking at the Julia source code we use `jl_dlfind` so we might look up symbols outside the Julia runtime?

vchuravy (commented):
One thought is around GC safety. Julia currently throws everything that is reachable from a native compilation into a global roots list (this has changed over time, I think 1.10 still had per CodeInstance rooting) so the thing I am wondering about is: Could we lose a root by going through this?

maleadt:
Another suggestion by @vchuravy: Try this out on AllocCheck.jl

src/relocation.jl
maleadt: > Why the 1-element array instead of a `Ref`?

Fable is convinced that `jl_value_ptr` on 1.14 can allocate temporary stack boxes and return its address, while `Any[]` forces the contents to be heap-allocated too.

```julia
@noinline jvp(x)  = UInt(ccall(:jl_value_ptr, Ptr{Cvoid}, (Any,), x))
@noinline slot(x) = (s = Any[x]; GC.@preserve s unsafe_load(Base.unsafe_convert(Ptr{UInt}, pointer(s))))

tag(a) = unsafe_load(Ptr{UInt}(a - 8)) & ~UInt(15)
const F64_TAG = jvp(Float64)                  # heap-interned DataType: fine either way
@noinline clobber(n) = n == 0 ? UInt(0) : clobber(n - 1) + hash(n)

function demo()
    for (name, a) in ("jl_value_ptr" => jvp(1.25), "Any[] slot" => slot(1.25))
        clobber(200)  # reuse the stack region the callee frame occupied
        println(rpad(name, 13), ": contents = ", rpad(unsafe_load(Ptr{Float64}(a)), 23),
                " valid type tag = ", tag(a) == F64_TAG)
    end
end
demo()
```

```
❯ jl +nightly mwe.jl
jl_value_ptr : contents = 6.92391611475455e-310   valid type tag = false
Any[] slot   : contents = 1.25                    valid type tag = true
```

src/optim.jl
maleadt: You mean make them all by-name references?

maleadt:
AllocCheck.jl integration: https://github.com/JuliaLang/AllocCheck.jl/pull/113
Uses a new relocation strategy, `:defer`, since AllocCheck.jl really does its own linking. This will be similar to what Enzyme.jl will have to do if it wants to benefit from this. I left the `:import` strategy in place, with an example user in the tests.

@vchuravy I think this is ready for another look.

maleadt:
> One thought is around GC safety. Julia currently throws everything that is reachable from a native compilation into a global roots list (this has changed over time, I think 1.10 still had per CodeInstance rooting) so the thing I am wondering about is: Could we lose a root by going through this?

I don't think so. `compile_method_instance` reads the GV table after `jl_create_native` returns, i.e. after root promotion. So every object wrapped in a `JuliaValueRef` is already permanently rooted, no?

EDIT: although I guess we could just make sure to perma-root, and then get rid of the `roots` mechanism in here? https://github.com/JuliaGPU/GPUCompiler.jl/pull/878/changes/86d4dac86ba455ae7c27a149d61608652263744b

maleadt:
Turns out Metal.jl doesn't support patching relocations like we do in CUDA (no support for something CuGlobal like), so I'm investigating a new strategy where we have a GOT-like table that's passed as part of the KernelState.

PR #879, opened 2026-07-14
[SPIRV] add workaround for khronos translator

simeonschaub:
works around https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3389

### DIFF
--- a/src/spirv.jl
+++ b/src/spirv.jl
@@ -353,6 +353,7 @@ function wrap_byval(@nospecialize(job::CompilerJob), mod::LLVM.Module, f::LLVM.F
     # NOTE: if we ever have legitimate uses of the old function, create a shim instead
     replace_function!(f, new_f)
 
+    # XXX: work around KhronosGroup/SPIRV-LLVM-Translator#3389
     if job.config.target.backend === :khronos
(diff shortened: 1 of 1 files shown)


src/spirv.jl
```diff
@@ -352,5 +352,12 @@ function wrap_byval(@nospecialize(job::CompilerJob), mod::LLVM.Module, f::LLVM.F
     # remove the old function
     # NOTE: if we ever have legitimate uses of the old function, create a shim instead
     replace_function!(f, new_f)
+
+    if job.config.target.backend === :khronos
```
maleadt, line 357:
```suggestion

    # XXX: work around KhronosGroup/SPIRV-LLVM-Translator#3389
    if job.config.target.backend === :khronos
```

maleadt (approved):
LGTM. Why are you still using the Khronos translator?

simeonschaub:
It was mostly motivated by us still running into issues with the LLVM backend like https://github.com/JuliaGPU/OpenCL.jl/issues/422, so I was just curious what things looked like on the Khronos translator side nowadays. Not sure we ever want to switch back, but it's nice to have another option if we ever do run into any major blockers with the LLVM backend

PR #888, opened 2026-07-22
relocate_gvs!: keep Type-valued globals as identity tokens

vchuravy:
## Problem

`relocate_gvs!` decides whether to materialize a global's value with:

```julia
if isbitstype(typeof(obj)) && sizeof(obj) > 0 && !(obj isa Bool)
```

`Union{}` slips through this guard. `typeof(Union{})` is `Core.TypeofBottom`, an isbits singleton, so `isbitstype(typeof(obj))` is `true` — but `sizeof(Union{})` **throws** (`The empty type does not have a definite size since it does not have instances.`) because the type is uninhabited.

This surfaces when differentiating ordinary code with [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) on Julia 1.12 (GPUCompiler v2 path): the primal module carries a `julia.constgv` global pointing at `Union{}`, and `relocate_gvs!` crashes while baking it in:

```
The empty type does not have a definite size since it does not have instances.
Stacktrace:
 [1] sizeof(x::Type)
 [2] relocate_gvs!(mod::LLVM.Module, gv_to_value::Dict{String, Ptr{Nothing}})
     @ GPUCompiler .../src/jlgen.jl:800
 [3] emit_llvm(job::GPUCompiler.CompilerJob)
```

## Fix

Exclude `obj isa Type` from materialization. Type objects — like any other non-isbits object — keep their host address as an identity token (baked `inttoptr`), which renders the module session-dependent, exactly as intended for objects whose identity matters.

## Test

Adds a `relocate_gvs!` unit test for a slot pointing at `Union{}`, asserting it is baked as an address rather than fed to `materialize_box!`. Verified the test crashes on `master` and passes with this change.

### DIFF
--- a/src/jlgen.jl
+++ b/src/jlgen.jl
@@ -796,8 +796,11 @@ function relocate_gvs!(mod::LLVM.Module, gv_to_value::Dict{String, Ptr{Cvoid}})
         val = nothing
         if init != C_NULL
             obj = Base.unsafe_pointer_to_objref(init)
-            # Zero-sized objects remain identity tokens.
-            if isbitstype(typeof(obj)) && sizeof(obj) > 0 && !(obj isa Bool)
+            # Zero-sized objects remain identity tokens. Type objects also stay
+            # identity tokens: some are isbits singletons (e.g. `Union{}`, whose
+            # `typeof` is the isbits singleton `Core.TypeofBottom`), but `sizeof`
+            # is undefined for uninhabited types and would throw here.
+            if isbitstype(typeof(obj)) && !(obj isa Type) && sizeof(obj) > 0 && !(obj isa Bool)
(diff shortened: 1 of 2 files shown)


topolarity:
Isn't this due to a deficiency in `Core.sizeof` trying to guess whether to query the object itself or the type?
```julia
julia> Core.sizeof(1)
8

julia> Core.sizeof(Int)
8
```

taking the `typeof` first should fix it:
```julia
julia> Core.sizeof(typeof(Int))
56

julia> Core.sizeof(typeof(Union{}))
0
```

maleadt:
This is not relevant anymore post https://github.com/JuliaGPU/GPUCompiler.jl/pull/878

PR #889, opened 2026-07-27
Disable materialize_box on uses_jl_runtime

vchuravy:
Noticed while updating Enzyme to GPUCompiler v2 and independently reproduced as a crash in AllocCheck.jl

x-ref: https://github.com/JuliaLang/AllocCheck.jl/pull/114


maleadt:
Huh, I missed this. Sorry for that.
Superseded by https://github.com/JuliaGPU/GPUCompiler.jl/pull/913

PR #896, opened 2026-08-17
Bump version.

simeonschaub:


maleadt:
Oh sorry I did not see this. I just tagged 2.2 with the relocatability changes (which should be nonbreaking).

PR #899, opened 2026-08-18
use `compilesig_invokes=false`

simeonschaub:
Otherwise, inlining rewrites `:invoke`s of `@noinline` vararg callees to their compileable (vararg-widened) signature, whereas inference infers and caches the fully-specialized callee, so codegen emits an unsupported dynamic call. In Julia 1.14, every bounds-check error for 3+-dimensional indexing goes through the vararg `Base._throw_boundserror_indices(A, i1, I...)` (JuliaLang/julia#61535), so without this, checked 3D `getindex` doesn't compile.

`compilesig_invokes=false` was added for external compilers exactly for this (JuliaLang/julia#46612): `:invoke`s then target the exact inferred specialization, whose CodeInstance is always present in the interpreter's cache.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -613,7 +613,13 @@ function inference_params(@nospecialize(job::CompilerJob))
 end
 
 # the optimization parameters to use when constructing the GPUInterpreter
-optimization_params(@nospecialize(job::CompilerJob)) = CC.OptimizationParams()
+#
+# Emit `:invoke`s targeting the exact inferred specialization instead of the
+# compileable (vararg-widened) signature: GPU back-ends cannot dispatch at run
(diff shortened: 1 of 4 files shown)


christiangnrd:
This broke Metal in Julia 1.12+

MWE:
```
using Metal; x = MtlArray{Bool}(undef, 2); x .= 1
```

simeonschaub:
Could you post the error message you get?

christiangnrd:
Ignore the non-`array` errors from this output https://buildkite.com/julialang/gpucompiler-dot-jl/builds/2677/table?jid=01a02553-6583-42f5-80ee-23e3125852c1&tab=output

PR #904, opened 2026-08-21
Metal: Demote unordered atomic loads and stores

simeonschaub:
The AGX back-end crashes on these with `XPC_ERROR_CONNECTION_INTERRUPTED`, because AIR has no LLVM-level atomic load/store instructions.

The `compilesig_invokes=false` change in #899 only surfaced this issue, since `Bool(::Real)` throws an `InexactError(:Bool, Bool, x)` through its vararg `@nospecialize` constructor, which previously was a dead dynamic call that `lower_throw!` just deleted. It's now a real `noinline` call whose body allocates the `Tuple{DataType,Int64}` with an unordered pointer store, breaking `MtlArray{Bool}(undef, 2) .= 1` on 1.12+.

Since there is no GC on GPUs, this should be legal, but please correct me otherwise

@christiangnrd Can you confirm whether this fixes the Metal issues?

### DIFF
--- a/src/metal.jl
+++ b/src/metal.jl
@@ -645,6 +645,23 @@ function finish_ir!(@nospecialize(job::CompilerJob{MetalCompilerTarget}), mod::L
     return functions(mod)[entry_fn]
 end
 
+# Julia marks heap-reference accesses `unordered` so raced pointer reads cannot produce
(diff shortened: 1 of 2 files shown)


christiangnrd:
This seems to have fixed things locally

Issue #906, opened 2026-08-21
#899 exposes allocating throw paths on malloc-less backends


michel2323:
GPUCompiler@v2.2.2 contains  #899, which sets `CC.OptimizationParams(; compilesig_invokes=false)`. With that, the `@noinline throw1(x) = throw(DomainError(x, ...))` closure inside `Base.Math.exponent` now gets codegen'd into the kernel, and `DomainError(x, msg)` needs a heap box → `gc_pool_alloc` → `gpu_malloc. gpu_malloc` is a backend-provided runtime symbol (`compile(:malloc, ...)` in GPUCompiler's `runtime.jl`), and oneAPI.jl doesn't provide one. Under 2.2.1, the throw path never reached the IR.

I'm working on a malloc for oneAPI.jl and pinned the GPUCompiler.jl version for now. From what I understand, the latter is needed for Julia 1.14.

christiangnrd:
See also https://github.com/JuliaGPU/GPUCompiler.jl/pull/904, https://github.com/JuliaGPU/Metal.jl/pull/928

@simeonschaub

simeonschaub:
Sorry, I didn't realize that change would cause so much breakage! Nothing showed up on CI, so I assumed it would be fine. I'd be ok with reverting #899 for now until we out how to properly handle these throw paths. I do think we want to eventually set `compilesig_invokes = false`, it's just that it exposes some paths that previously didn't really get handled

maleadt:
I don't think we can avoid this, see https://github.com/JuliaGPU/GPUCompiler.jl/pull/908#issuecomment-5557864303. However, all back-ends should have a device-allocator now, so I think we can close this.

PR #908, opened 2026-08-22
Preserve inferred specializations in invokes

maleadt:
Set `compilesig_invokes=false` so non-inlined calls retain their inferred specialization instead of targeting a widened compilation signature that may not be cached. This avoids unsupported dynamic invokes, including the multidimensional bounds-error paths introduced in Julia 1.14.

For backends without a device heap, provide a null-returning allocator when their runtime module does not define `malloc`. Valid inputs can then compile and execute even when an error path needs to box an exception; attempted allocations follow the existing out-of-memory path. Backend-provided allocators retain precedence. This addresses #906.

The branch includes the Metal atomic-lowering fix from #904 through its base. Add a Bool-conversion regression with an opaque allocator so LLVM cannot erase the heap-reference stores before AIR lowering, as the usual constant-null test allocator does.

The existing `optimization_params(job)` hook remains available for backend overrides. Unlike `infer_compilation_signature`, which also infers widened signatures, disabling `compilesig_invokes` preserves the specialization selected by inference.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -621,7 +621,13 @@ function inference_params(@nospecialize(job::CompilerJob))
 end
 
 # the optimization parameters to use when constructing the GPUInterpreter
-optimization_params(@nospecialize(job::CompilerJob)) = CC.OptimizationParams()
+#
+# Keep invokes specialized to the inferred argument types. Widening to compilation
+# signatures can lose type information or target code absent from our cache.
+# `infer_compilation_signature` would infer those wider signatures too; it does not
+# preserve specialization. Back-ends can override this hook to change the policy.
+optimization_params(@nospecialize(job::CompilerJob)) =
+    CC.OptimizationParams(; compilesig_invokes=false)
 
 # how much debuginfo to emit
 function llvm_debug_info(@nospecialize(job::CompilerJob))
--- a/test/gcn.jl
+++ b/test/gcn.jl
@@ -520,9 +520,10 @@ end
 
     @test @filecheck begin
         @check_label "define void @{{(julia|j)_kernel_[0-9]+}}"
-        # box: a jl_box_float32 call <1.14; 1.14+ inlines it into the devirt'd ctor
-        @check cond=(VERSION < v"1.14-")  "jl_box_float32"
-        @check cond=(VERSION >= v"1.14-") "gpu_gc_pool_alloc"
+        # 1.10 boxes through jl_box_float32; 1.11+ specializes the constructor
+        # and boxes through the GC pool allocator instead
+        @check cond=(VERSION < v"1.11-")  "jl_box_float32"
+        @check cond=(VERSION >= v"1.11-") "gpu_gc_pool_alloc"
         GCN.code_llvm(mod.kernel, Tuple{Float32,Ptr{Float32}}; dump_module=true)
     end
     GCN.code_native(devnull, mod.kernel, Tuple{Float32,Ptr{Float32}})
--- a/test/metal.jl
+++ b/test/metal.jl
@@ -1557,6 +1557,38 @@ end
     end
 end
 
+@testset "Bool conversion exception allocation" begin
+    mod = @eval module $(gensym())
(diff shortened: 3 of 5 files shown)


src/interface.jl
```diff
@@ -615,7 +615,13 @@ function inference_params(@nospecialize(job::CompilerJob))
...
 
 # the optimization parameters to use when constructing the GPUInterpreter
-optimization_params(@nospecialize(job::CompilerJob)) = CC.OptimizationParams()
+#
+# Emit `:invoke`s targeting the exact inferred specialization instead of the
+# compileable (vararg-widened) signature: GPU back-ends cannot dispatch at run
+# time, and widened signatures may have no inferred code in our cache (e.g.
+# `Base._throw_boundserror_indices(A, i1, I...)` from bounds checks).
+optimization_params(@nospecialize(job::CompilerJob)) =
+    CC.OptimizationParams(; compilesig_invokes=false)
```
vchuravy, line 630: We might have to make this configurable And how does this differ from `Core.Compiler.infer_compilation_signature`?

maleadt: `optimization_params` is already an overridable hook, no?

And I'm not familiar with the other flag. According to Astra:
> Difference from `infer_compilation_signature`: that flag asks inference to also infer the widened compilation signature. `compilesig_invokes=false` instead prevents the optimizer from redirecting invokes to that wider specialization. Inferring the wider signature can address a missing cache entry, but does not preserve the concrete argument types needed for static dispatch.

maleadt:
Why this causes oneAPI.jl failures:

> Before, with widened invokes. The optimizer rewrites the invoke target to the constructor's compilation signature. Because of @nospecialize, that is DomainError(::Any, ::Any). Codegen must hand that callee boxed arguments, so it boxes x at the call site, inside throw1:
> 
> ```
> %box = julia.gc_alloc_obj(8 bytes, Float64)
> store double %x, %box
> %exc = call DomainError(::Any, ::Any)(%box, %msg)   ; or a jl_invoke if uncached
> call jl_throw(%exc)
> ```
> 
> lower_throw! then replaces the jl_throw with report, signal, and trap, and its hack erases the throw's argument when that argument is an instruction with no other uses. That deletes the constructor call. Now %box has no consumer. Since julia.gc_alloc_obj is an allocation intrinsic whose result nobody reads, LLVM removes it together with the store, before the GC-lowering pass would have turned it into gc_pool_alloc. No gpu_malloc ever appears. The hack's own comment says it was written for exactly this shape: "throwing objects with @nospecialize constructors".
> 
> After, with specialized invokes. The invoke keeps the inferred target, the specialization for (Float64, String). That callee takes the double unboxed and returns the struct through an sret pointer, because DomainError is an immutable with two pointer fields. The boxing now happens inside the constructor body, where val is stored into an Any field:
> 
> ```
> ; in throw1
> call DomainError_spec(sret %tmp, double %x, %msg)
> %obj = julia.gc_alloc_obj(16 bytes, DomainError)
> memcpy(%obj, %tmp, 16)
> call jl_throw(%obj)
> 
> ; in DomainError_spec
> %box = julia.gc_alloc_obj(8 bytes, Float64)     ; the argument box
> store double %val, %box
> store %box, %sret[0]
> ```
> 
> Two things defeat the lowering here. The throw's argument %obj still has a use, the memcpy, so the hack does not erase it. And the constructor call is not connected to the throw through any SSA use at all: its output flows through memory. Even if the outer object were deleted, the box inside the callee is stored into the sret result, so it escapes and LLVM cannot drop it. GC lowering then turns it into gc_pool_alloc, which calls malloc, which on a back-end without one is an unresolved gpu_malloc.

So I think we need to wait until oneAPI.jl gets a device allocator.

Metal.jl was already fixed in #904 and JuliaGPU/Metal.jl#928

maleadt:
oneAPI.jl has a device allocator now, so this should be good to go. We just shouldn't tag before oneAPI.jl has gotten a new release.

Issue #909, opened 2026-08-26
Concrete isbits type with `missing` as a type parameter causes error during mangling


joachimbrand:
When writing a kernel that does a trivial operation on a custom isbits type that has a type parameter that is deliberately set to `missing` I get an error from GPUCompiler. The code runs fine on a CPU and works with a concrete isbits type, so I consider it a bug.

I'll submit a PR as well that will fix the bug.

## Error message
The full error message is here:
```
ERROR: TypeError: non-boolean (Missing) used in boolean context
Stacktrace:
  [1] mangle_param(t::Any, substitutions::Vector{Any}, top::Bool)
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/mangling.jl:37
  [2] mangle_param(t::Any, substitutions::Vector{Any}, top::Bool)
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/mangling.jl:94
  [3] mangle_param(t::Any, substitutions::Vector{Any})
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/mangling.jl:37
  [4] mangle_param(t::Any, substitutions::Vector{Any}, top::Bool)
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/mangling.jl:94
  [5] mangle_sig(sig::Any)
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/mangling.jl:189
  [6] relocation_namespace
    @ ~/git/code/GPUCompiler.jl/src/relocation.jl:334 [inlined]
  [7] collect_julia_value_relocations!(job::GPUCompiler.CompilerJob, mod::LLVM.Module, gv_to_value::Dict{String, Ptr{…}})
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/relocation.jl:346
  [8] irgen(job::GPUCompiler.CompilerJob)
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/irgen.jl:5
  [9] emit_llvm(job::GPUCompiler.CompilerJob; resolve_relocations::Bool)
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/driver.jl:188
 [10] compile_unhooked(output::Symbol, job::GPUCompiler.CompilerJob; resolve_relocations::Bool)
    @ GPUCompiler ~/git/code/GPUCompiler.jl/src/driver.jl:80
 [11] compile_unhooked
[...]

PR #910, opened 2026-08-26
Use identity comparison during type mangling

joachimbrand:
Fixes issue #909 

The proposed change is to use an identity instead of value comparison when checking for absent type parameters during type mangling.

A test was added that confirms that `missing` is a valid type parameter.


maleadt (approved):

joachimbrand:
Thanks 😄

PR #919, opened 2026-09-03
Handle allocation failure when boxing

maleadt:
`Runtime.box` wrote the type tag and payload through every pointer returned by `malloc`. OpenCL has no device allocator and returns `C_NULL`; after inlining, back-end optimizers can treat those stores as undefined and remove the throwing path, including `signal_exception`.

Use the same failure path as `gc_pool_alloc`: report OOM and throw `OutOfMemoryError` before writing. This prevents an invalid Julia object from escaping and gives optimization a real exceptional path. Cover both allocator-less SPIR-V back-ends and verify the OOM and original exception paths.

### DIFF
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -243,6 +243,11 @@ end
     quote
         ptr = malloc($(Csize_t(allocsz)))
 
+        if ptr == C_NULL
+            report_oom($(Csize_t(allocsz)))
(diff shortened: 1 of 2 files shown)


maleadt:
This is a prerequisite for improved exception handling in OpenCL.jl (without having to add a device-side allocator).

PR #920, opened 2026-09-04
Pass runtime strings in the target's global address space

maleadt:
Exception names and backtrace strings are emitted in the target's global address space, but the runtime declarations used `Ptr{Cchar}` and cast them to the default address space. For SPIR-V, that does not describe the cross-workgroup storage containing the strings.

Represent runtime string arguments with a placeholder resolved per compiler job. Keep `Ptr{Cchar}` as the default and derive SPIR-V's pointer type from the data layout's global address space, avoiding an address-space cast. Cover both typed and opaque LLVM pointers and translate the result through both SPIR-V back-ends.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -113,7 +113,7 @@ else
 end
 
 const CONFIG_KWARGS = [:kernel, :name, :entry_abi, :always_inline, :opt_level,
-                       :libraries, :optimize, :cleanup, :validate, :strip]
+                       :debug_level, :libraries, :optimize, :cleanup, :validate, :strip]
 
 """
     CompilerConfig(target, params; kernel=true, entry_abi=:specfunc, name=nothing,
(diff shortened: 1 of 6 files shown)


maleadt:
I don't particularly like the new interface though, so I'm also going to explore getting this out of LLVM.

vchuravy:
I thought we had already tried to fix this before?

vchuravy:
https://github.com/JuliaGPU/OpenCL.jl/pull/302/changes

We have `globalstring_ptr!(builder, String(fmt); addrspace=AS.Constant)`

Edit: Ah I see the issue, we are emitting in AS.Constant, but your intrinsics would like to see AS 0

maleadt:
> Edit: Ah I see the issue, we are emitting in AS.Constant, but your intrinsics would like to see AS 0

Yes. But in the latest version, I made it so that we respecialize the runtime functions on a different type, so they are really accepting AS1 now, i.e., what `globalstring_ptr!` produces.

PR #924, opened 2026-09-06
Demote unordered atomic accesses for SPIR-V targets

maleadt:
Julia marks loads and stores of heap references `unordered`, so that a read racing with the GC or with another thread's write cannot observe a torn pointer. There is no device GC for these orderings to protect against, but not every back-end can express them: SPIR-V's OpAtomicLoad/OpAtomicStore only take scalar integer or floating-point operands, so the Khronos translator turns an `unordered` load of a pointer into an invalid pointer-typed atomic ("AtomicLoad: expected Result Type to be integer or float scalar type" from spirv-val; Intel's compiler fails on an undefined `__spirv_AtomicLoad(long**, int, int)`). The LLVM SPIR-V back-end already lowers such loads to plain OpLoads. With oneAPI.jl gaining a device allocator, kernels that read boxed fields now reach this on Julia 1.12+.

Move the Metal target's `demote_unordered_atomics!` next to the other shared post-optimization rewrites and run it from the SPIR-V target's `finish_ir!` too. The end-to-end test compiles an `unordered` pointer load and store through both SPIR-V code paths with validation enabled.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "GPUCompiler"
 uuid = "61eb1bfa-7361-4325-ad38-22787b887f55"
-version = "2.5.3"
+version = "2.5.4"
 authors = ["Tim Besard <tim.besard@gmail.com>"]
 
 [workspace]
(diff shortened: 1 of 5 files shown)


maleadt:
CI failure unrelated, looks like Julia+Asserts has regressed and fails to precompile e.g. Parsers.jl
EDIT: https://github.com/JuliaLang/julia/pull/63036

PR #928, opened 2026-09-09
Internalize non-kernel deferred codegen entrypoints

vchuravy:
Motivated by JuliaGPU/KernelAbstractions.jl#736.

### Problem

Deferred codegen entrypoints (notably the wrappers Enzyme generates) were held externally live across GPUCompiler's `InternalizePass` so that linking could resolve them. Once linked and `alwaysinline`d they are dead, but external linkage kept `GlobalDCEPass` from dropping them, so e.g. the SPIR-V back-end still had to translate functions it cannot express.

### Change

Only deferred jobs that are kernels themselves (e.g. child kernels for CUDA dynamic parallelism, which inherit the parent's config) are entrypoints in their own right. The driver now classifies the entrypoints once after deferred codegen, and uses that set both for internalization and for the per-entrypoint `finish_ir!` loop. Other deferred functions (Enzyme registers its jobs with `kernel=false`) are internalized like any other function, and dropped by the existing clean-up `GlobalDCEPass` once inlined. No back-end hooks are touched, so the `finish_ir!` override in the KernelAbstractions PR should no longer be necessary.

### Tests

- The mock Enzyme helper gains an opt-in `always_inline` flag on `Enzyme.deferred_codegen`, marking the generated function `alwaysinline` like Enzyme's wrappers.
- Native and PTX "Mock Enzyme" tests now assert the deferred child is `define internal`, and that it is gone entirely when `alwaysinline`. The PTX test previously compiled for the native target; it now compiles an actual PTX kernel.
- New SPIR-V "deferred codegen" testset (both back-ends): the `alwaysinline` child is absent from the optimized module, and the translated SPIR-V contains exactly one `OpFunction`. Without the driver change this fails on both back-ends, as the external definition is left behind.

Note that the reflection `code_native` sets `only_entry` unless `dump_module=true`, which silently skips deferred codegen; the SPIR-V test passes `dump_module=true` for that reason.

Tested locally with Julia 1.10 and 1.12 (`native`, `ptx`, `spirv`, `utils`, `gcn`, `metal`).

https://claude.ai/code/session_01Jap1wwekUHPKo3iET2dKhn

### DIFF
--- a/src/driver.jl
+++ b/src/driver.jl
@@ -292,6 +292,13 @@ const __llvm_initialized = Ref(false)
         erase!(dyn_marker)
     end
 
+    # only deferred jobs that are kernels themselves (e.g. child kernels for dynamic
+    # parallelism) are entrypoints in their own right. other deferred functions (e.g. the
+    # wrappers Enzyme generates) are only called from within this module, so they should
+    # be internalized like any other function and dropped once inlined, rather than kept
+    # around for back-ends that cannot express their signatures (e.g. SPIR-V).
+    entrypoints = filter(((job′, _),) -> job′ === job || job′.config.kernel, jobs)
+
     if job.config.toplevel && job.config.libraries
         # load the runtime outside of a timing block (because it recurses into the compiler)
         if !uses_julia_runtime(job)
@@ -320,7 +327,7 @@ const __llvm_initialized = Ref(false)
             # mark everything internal except for entrypoints and any exported
             # global variables. this makes sure that the optimizer can, e.g.,
             # rewrite function signatures.
-            preserved_gvs = collect(values(jobs))
+            preserved_gvs = collect(values(entrypoints))
             for gvar in globals(ir)
                 if linkage(gvar) == LLVM.API.LLVMExternalLinkage
                     push!(preserved_gvs, LLVM.name(gvar))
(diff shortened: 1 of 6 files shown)


maleadt:
Ugh, this is becoming one hell of an ugly interface.

src/driver.jl
```diff
@@ -292,6 +292,13 @@ const __llvm_initialized = Ref(false)
         erase!(dyn_marker)
     end
 
+    # only deferred jobs that are kernels themselves (e.g. child kernels for dynamic
+    # parallelism) are entrypoints in their own right. other deferred functions (e.g. the
+    # wrappers Enzyme generates) are only called from within this module, so they should
+    # be internalized like any other function and dropped once inlined, rather than kept
+    # around for back-ends that cannot express their signatures (e.g. SPIR-V).
+    entrypoints = filter(((job′, _),) -> job′ === job || job′.config.kernel, jobs)
```
maleadt, line 300: `jobs` is being used for too many things here. It was only intended to be for deduplicating compilation, but now it's also for preserving externally visible symbols, and scheduling finalization... That makes this code really hard to read.

I'm fine with the bugfix, but we should really replace this with separate mechanisms at some point.

vchuravy:
Not disagreeing, and I am in a catch-22 of my own making. KA 0.10 should keep Enzyme support...

I would like to re-engineer the GPU integration that Enzyme currently has. Maybe Claude has more follow-through than I do on #582

